PR #522, opened 2018-08-29
Update config.yml

dexhorthy:
What I Did
------------

Make tests that integrate with pg.replicated.com opt-in

How I Did it
------------

Check if the required token is set, else skip

How to verify it
------------

Build from a fork that does not have the variable set

Description for the Changelog
------------

Ship integration testing in CI no longer requires a token to access pg.replicated.com. These tests will still be run if the token is set in the build environment.


laverya:
I think I'd rather split the integration task into two - one for `init_app` or anything else that ends up requiring an env var, one for all the rest

That way we can keep `integration` a required test, and just ignore the failing `integration-secrets` test

That sort of thing

dexhorthy:
I think we'd rather have this fail silently, then fail explicitly but not block merging. The latter encourages merging in the face of ❌

PR #546, opened 2018-08-31
Retry fetching upstream

### LINKED ISSUE #496: ship init fails after 1 attempt of downloading chart
### Expected behavior
If the upstream returns a timeout or an error, ship init should retry.

### Actual behavior
Ship init fails after the first upstream error. This occasionally causes test failures because the tests depend on upstream helm charts to render and run.

Rob0h:
What I Did
------------
- Fix #496 

How I Did it
------------
- Add a flag of `retries` which defaults to 3
- Retries fetching via go-getter or github

How to verify it
------------
- Run `ship` against a bad url

Description for the Changelog
------------
Retry fetching upstream

------------
🛶

### DIFF
--- a/pkg/specs/apptype/determine_type.go
+++ b/pkg/specs/apptype/determine_type.go
@@ -109,7 +109,8 @@ func (r *inspector) determineTypeFromContents(
 
 			var retryError error
 			retries := r.viper.GetInt("retries")
-			for idx := 1; idx <= retries; idx++ {
+			hasSucceeded := false
+			for idx := 1; idx <= retries && !hasSucceeded; idx++ {
 				debug.Log("event", "retry.getFiles", "attempt", idx)
 				r.ui.Info(fmt.Sprintf("Retrying to retrieve upstream %s ...", upstream))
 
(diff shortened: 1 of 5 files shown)


pkg/specs/apptype/determine_type.go
```diff
@@ -101,6 +104,26 @@ func (r *inspector) determineTypeFromContents(
...
+				r.ui.Info(fmt.Sprintf("Retrying to retrieve upstream %s ...", upstream))
+
+				time.Sleep(time.Second * 5)
+				retryError = fetcher.GetFiles(ctx, upstream, savePath)
+
+				if retryError != nil {
+					r.ui.Info(fmt.Sprintf("Retry attempt %v out of %v to fetch upstream failed", idx, retries))
+					level.Error(r.logger).Log("event", "getFiles", "err", retryError)
+				} else {
+					return "", "", nil
```
laverya: I think this block needs some reworking
Correct flow should be return error after retries run out (including if 0 retries) and continue on into the next code on success

pkg/specs/apptype/determine_type.go
```diff
@@ -109,7 +109,12 @@ func (r *inspector) determineTypeFromContents(
 
 			var retryError error
 			retries := r.viper.GetInt("retries")
+			hasSucceeded := false
 			for idx := 1; idx <= retries; idx++ {
+				if hasSucceeded {
+					break
+				}
```
laverya: `for idx := 1; idx <= retries && !hasSucceeded; idx++ {`?

Issue #527, opened 2018-08-30
Ship init with replicated apps should store the app details in state.json


laverya:
### Expected behavior
Customer ID, Installation ID, release version and content hash should all be stored in `.ship/state.json` as with non-replicated apps.

### Actual behavior
Only config items are stored in `.ship/state.json`.

### Information

- Ship version: master as of https://github.com/replicatedhq/ship/commit/ebb027556d4c3fbaa7e407491188e42f7a6938da
- Command line run: `ship init staging.replicated.app/some-cool-ci-tool?installation_id=<INSTALLATION_ID>&customer_id=<CUSTOMER_ID>`

### Steps to reproduce the behavior

1. Look at the expected [.ship/state.json](https://github.com/replicatedhq/ship/blob/ebb027556d4c3fbaa7e407491188e42f7a6938da/integration/init_app/github-template-func/expected/.ship/state.json) files in the init_app integration tests.

PR #558, opened 2018-09-14
Serialize metadata to state

Rob0h:
What I Did
------------
- Write version, icon, name to state

How I Did it
------------
- Add `SerializeMetadata` to `Manager`

How to verify it
------------
- Test

Description for the Changelog
------------
Serialize metadata to state

------------
🛶


laverya:
`state.json` is checked in the integration tests, so that's going to have to be updated (which is going to be a LOT of changes). Same goes for some of the unit tests, it looks like.

It would probably also be a good idea to fix #527 as part of this PR, since that will also involve modifying `state.json`. Better to do it once, right? (we don't currently store semver for replicated apps in state.json, and we need to)

PR #570, opened 2018-09-19
Add kustomize build functionality, build by default

dexhorthy:
What I Did
------------

Update kustomize lifeycle step to build if `dest` is set. The default raw/helm lifecycles now build by default with a dest of `rendered.yaml`. Renamed some fields that could cause breaking changes in ship.yaml apps, but usage of kustomize in lifecycle is low, and we've applied sane defaults.

How I Did it
------------

- rename some fields on the `Kustomize` lifecycle step. `overlay` is now the path to the overlay we create, defaulting to `overlays/ship`. This field used to be called `dest`,
-`dest` is now where the `kustomize build` output will get written. `dest` is optional. If unset we will not do a `kustomize build`.
- If `Dest` is set, we try to get a `render` step for the lifecycle, and use its `Root` property to prefix the `kustomize build` destination (this has no effect in ship init, but may affect using a kustomize step in ship.yaml / replicated.app)
- adjust default lifecycle to build to `rendered.yaml`
- update instructions to use `kubectl apply -f rendered.yaml` instead of the previous `kustomize build overlays/ship | kubectl apply -f -` command.

- update integration tests to check for rendered.yaml
- add `make clean-integration` to clean all test runs
- add support for `SHIP_INTEGRATION_SKIP_CLEANUP_ALL` in integration tests
- change the stable-mysql to use a chart that has deterministic output (`stable/factorio`). Almost every database chart generates a random password, but this one hardcodes a placeholder.
-

Other stuff
---------

- fix a nil pointer panic in replicatedapp/resolver.go
- fix a file handle leak when unpacking github files
- dont include Chart.yaml or values.yaml in `kustomization.yaml#resources`. This is no longer needed for the test I was fixing, but feels like a good idea regardless.

How to verify it
------------

```
ship init ...
diff <(cat rendered.yaml) <(kustomize build overlays/ship)
```

Diff should be empty.

Description for the Changelog
------------

- By default, `ship init` now runs the equivalent of `kustomize build` during execution. Instructions have been updated to suggest a `kubectl apply` instead of the previous `kustomize build ... | kubectl apply -f -` command.
- Fix a file handle leak when unpacking artifacts from github repos


pkg/api/lifecycle_test.go
```diff
@@ -55,18 +55,37 @@ lifecycle:
...
+			name: "kustomize with dest",
+
+			yaml: `
+---
+lifecycle:
+  v1:
+    - kustomize:
+         base: k8s/
+         dest: rendered.yaml
+         overlay: overlays/ship`,
```
dexhorthy: example of changed fields

Makefile
```diff
@@ -309,6 +309,9 @@ clean-ship:
 	rm -rf base/
 	rm -rf .ship/
 
+clean-integration:
+	rm -rf integration/*/*/_test_*
```
laverya: 👍

integration/init/git-root-directory/metadata.yaml
```diff
@@ -1,3 +1,3 @@
-upstream: "https://github.com/replicatedhq/test-charts/tree/ad1e78d13c33fae7a7ce22ed19920945ceea23e9/"
+upstream: "https://github.com/replicatedhq/test-chart-root-dir/tree/507feecae588c958ebe82bcf701b8be63f34ac9b/"
 args: []
-skip_cleanup: false
+skip_cleanup: true
```
laverya: You mean to leave this in?

pkg/api/lifecycle.go
```diff
@@ -94,8 +94,17 @@ func (t *Terraform) ShortName() string   { return "terraform" }
 // It does not take a kustomization.yml, rather it will generate one in the .ship/ folder
 type Kustomize struct {
 	StepShared `json:",inline" yaml:",inline" hcl:",inline"`
-	BasePath   string `json:"base_path,omitempty" yaml:"base_path,omitempty" hcl:"base_path,omitempty"`
+	Base       string `json:"base,omitempty" yaml:"base,omitempty" hcl:"base,omitempty"`
 	Dest       string `json:"dest,omitempty" yaml:"dest,omitempty" hcl:"dest,omitempty"`
+	Overlay    string `json:"overlay,omitempty" yaml:"overlay,omitempty" hcl:"overlay,omitempty"`
```
laverya: I think you might have to regenerate the docs (and update mutations.json)

pkg/lifecycle/kustomize/kustomizer.go
```diff
@@ -83,26 +80,58 @@ func (l *daemonkustomizer) Execute(ctx context.Context, release *api.Release, st
...
+			return errors.Wrap(err, "build overlay")
+		}
+	}
+
 	return nil
 }
 
+// hack -- get the root path off a render step to tell if we should prefix kustomize outputs
+func (l *Kustomizer) getPotentiallyChrootedFs(release *api.Release) (afero.Afero, error) {
+	renderRoot := "installer"
```
Rob0h: `constants.InstallerPrefixPath` ?

integration/init/git-root-directory/metadata.yaml
dexhorthy: nope

PR #571, opened 2018-09-19
Scope body, reset to ship-init class, scope all other css to #ship-init-component

10Dimensional:
What I Did
------------
Added a `ship-init` class on mount to body tag to scope reset and base styles to ship-init app.

How I Did it
------------
Changed the Ship.jsx component to a class, added `componentDidMount` and `componentDidMount` lifecycle methods to handle class addition. Moved all reset and base styles underneath `body.ship-init` and scoped all component styles to `#ship-init-component`.

How to verify it
------------
Run the web app in dev mode

Description for the Changelog
------------
Namespaced ship-init UI CSS

------------
Boat

### DIFF
--- a/web/init/src/Ship.jsx
+++ b/web/init/src/Ship.jsx
@@ -6,16 +6,30 @@ import { configureStore } from "./redux";
 import PropTypes from "prop-types";
 
 import "./scss/index.scss";
+const bodyClass = "ship-init";
 
-export const Ship = ({ apiEndpoint }) => (
-  <div id="ship-init-component">
-    <Provider store={configureStore(apiEndpoint)}>
-      <AppWrapper>
-        <RouteDecider />
-      </AppWrapper>
-    </Provider>
-  </div>
-);
+export class Ship extends React.Component {
+  componentDidMount() {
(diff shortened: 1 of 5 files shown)


web/init/src/Ship.jsx
```diff
@@ -6,16 +6,30 @@ import { configureStore } from "./redux";
...
-    <Provider store={configureStore(apiEndpoint)}>
-      <AppWrapper>
-        <RouteDecider />
-      </AppWrapper>
-    </Provider>
-  </div>
-);
+export class Ship extends React.Component {
+  componentDidMount() {
+    document.body.classList.add(bodyClass);
```
jadengore: Just leaving a note here that React components should avoid modifying the global document or using unexpected global styles that could conflict with other components.

For now this is fine but it will have to be a temporary fix.

jadengore:
@10Dimensional Also please bump the version number of `@replicatedhq/ship-init` so we deploy the new version!

web/init/src/Ship.jsx
10Dimensional: Issue: https://github.com/replicatedhq/ship/issues/572

PR #582, opened 2018-09-21
Add basePath prop to Ship init component 

jadengore:
What I Did
------------
Adds a `basePath` prop to the `Ship` component for customizing the base route of the internal router.

Bumps version of component to `v1.1.0` due to the addition of a prop.

How I Did it
------------
- Update router to accept a base path parameter for using the routing system external from the Ship binary
- Fix regression where skipping kustomize step would fail

How to verify it
------------
Our existing E2E test should continue to pass, since I have provided the `basePath` we use for the project (`/`).

Description for the Changelog
------------
- Add basePath prop to Ship init component 

------------
⛵️

### DIFF
--- a/web/init/src/components/shared/RouteDecider.jsx
+++ b/web/init/src/components/shared/RouteDecider.jsx
@@ -1,4 +1,5 @@
 import React from "react";
+import PropTypes from "prop-types";
 import { BrowserRouter, Route, Switch } from "react-router-dom";
 import isEmpty from "lodash/isEmpty";
 import NavBar from "../../containers/Navbar";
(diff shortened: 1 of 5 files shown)


web/init/src/components/shared/RouteDecider.jsx
```diff
@@ -8,18 +9,40 @@ import StepNumbers from "./StepNumbers";
...
-          this.props.getHelmChartMetadata();
+    const {
+      basePath,
+      routes,
+      getHelmChartMetadata,
+    } = this.props
+    if (routes !== lastProps.routes && routes.length) {
+      for (let i = 0; i < routes.length; i++) {
+        if (routes[i].phase.includes("helm")) {
+          getHelmChartMetadata();
```
dexhorthy: is this just reformatting, right?

PR #634, opened 2018-10-10
load resources from state into UI tree

dexhorthy:
What I Did
------------

Load `Resources` from state.json into the UI tree alongside patched

How I Did it
------------

Copy paste a lot of code around patches in `LoadTree`. There's some dupe emerging, but I feel that for now its the right kind of duplication, I dont think  we ought to be trying to consolidate an abstraction around patches vs resources.

How to verify it
------------

```
ship init ...
vim state.json # add resources
ship update --headed # see resources in tree
```

Description for the Changelog
------------

Load resources from state into UI tree


pkg/specs/chart.go
```diff
@@ -157,10 +156,9 @@ func (r *Resolver) resolveMetadata(ctx context.Context, upstream, localPath stri
 	}
 
 	if util.IsGithubURL(upstream) {
-		githubClient := githubclient.NewGithubClient(r.FS, r.Logger)
-		releaseNotes, err := githubClient.ResolveReleaseNotes(ctx, upstream)
+		releaseNotes, err := r.GitHubReleaseNotesFetcher.ResolveReleaseNotes(ctx, upstream)
```
dexhorthy: this is to keep a test from making a call out to github.com every time

PR #637, opened 2018-10-11
Update overlay/patch terminology

dexhorthy:
What I Did
------------

A few copy updates to unsure we're using correct overlay/patch terminology

How I Did it
------------

words

How to verify it
------------

Run through ship init, compare to https://github.com/kubernetes-sigs/kustomize/blob/master/docs/glossary.md

Description for the Changelog
------------

Update overlay/patch terminology in kustomize steps.

:canoe:


fshot:
This all looks good to me.  Good call tightening up the terminology now, so we don't end up muddying the waters and training our users to use wrong terms.

PR #639, opened 2018-10-15
updates for add resources workflow

dexhorthy:
What I Did
------------

Improvements for managing non-patch resources during Kustomize step.

How I Did it
------------

- In addition to `/` (base) and `overlays` (patches), include another `filetree.Node` in the return from `GET /api/v1/navcyle/step/kustomize`, called `resources`
- Include all added `resources` from state.json in this node.
- Add endpoint to delete a resource, refactoring out core delete logic into helper

How to verify it
------------

With ship running, add `resources` to `state.json`, try out `curl`s to `GET /api/v1/navcycle/step/kustomize` and `DELETE /api/v1/kustomize/resource/<filename>`

Description for the Changelog
------------

Improved API support for deleting resources.

:canoe:

### DIFF
--- a/pkg/filetree/loader.go
+++ b/pkg/filetree/loader.go
@@ -98,6 +98,10 @@ func (a *aferoLoader) LoadTree(root string) (*Node, error) {
 	}
 
 	populatedBase, err := a.loadTree(fs, rootNode, files)
+	if err != nil {
+		return nil, errors.Wrap(err, "load tree")
+	}
+
 	populatedPatches := a.loadOverlayTree(patchesRootNode, a.patches)
 	populatedResources := a.loadOverlayTree(resourceRootNode, a.resources)
 
@@ -115,7 +119,7 @@ func (a *aferoLoader) LoadTree(root string) (*Node, error) {
 		Path:     "/",
 		Name:     "/",
 		Children: children,
-	}, errors.Wrap(err, "load tree")
+	}, nil
 }
 
 // todo move this to a new struct or something
--- a/pkg/filetree/loader.go
+++ b/pkg/filetree/loader.go
@@ -19,7 +19,8 @@ import (
 
 const (
 	CustomResourceDefinition = "CustomResourceDefinition"
-	OverlaysFolder           = "overlays"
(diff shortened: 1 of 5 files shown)


pkg/filetree/loader.go
```diff
@@ -85,18 +86,29 @@ func (a *aferoLoader) LoadTree(root string) (*Node, error) {
...
+		Name:     PatchesFolder,
 		Children: []Node{},
 	}
+	resourceRootNode := Node{
+		Path:     "/",
+		Name:     ResourcesFolder,
+		Children: []Node{},
+	}
+
+	populatedBase, err := a.loadTree(fs, rootNode, files)
```
Rob0h: what happens to this error 😢

pkg/lifecycle/daemon/routes_navcycle_kustomize.go
```diff
@@ -298,57 +298,81 @@ func (d *NavcycleRoutes) createOrMergePatch(c *gin.Context) {
...
-		return
+		return errors.New("current kustomize empty")
 	}
 
 	shipOverlay := kustomize.Ship()
-	if len(shipOverlay.Patches) == 0 {
-		level.Error(d.Logger).Log("event", "empty ship overlay")
-		c.AbortWithError(http.StatusBadRequest, errors.New("bad delete request"))
-		return
+	files := filesMap(shipOverlay)
```
dexhorthy: I'm gonna rename `filesMap` to `getFiles` for clarity

pkg/filetree/loader.go
dexhorthy: oh good call

dexhorthy: looks like it was never being handled?

dexhorthy: wait - no -it happens up here https://github.com/replicatedhq/ship/pull/639/files#diff-054ba8f349d9b1eadd46df29c7372a83L106

dexhorthy: but that gap is getting pretty wide and we probably ought to just `!= nil` it instead

PR #640, opened 2018-10-15
Add customization of rendered destination path to DefaultHelmRelease 

jadengore:
What I Did
------------
This change passes the name of the Helm chart as the rendered destination name.
Before, the final rendered chart filename was named `rendered.yaml`.
With this change, the final rendered chart filename is named `<chartname>.yaml`.

This change only applies to Helm charts. All Kubernetes/Replicated apps will continue to put the final rendered YAML in `rendered.yaml`.

How I Did it
------------
- Refactor Helm/k8s shared logic
  * Split it out a bit more, since there are some key differences between how the release YAML is resolved on these two code paths.
- Update unit/integration tests

How to verify it
------------
Unit/integration tests verify this change, I have renamed the final rendered Helm charts to their respective Helm chart names.

Description for the Changelog
------------
- Add customization of rendered destination to DefaultHelmRelease 

------------


jadengore:
[ch15204]

PR #641, opened 2018-10-16
Feature/add resource

10Dimensional:
What I Did
------------
Updated the ship init UI flow to handle new resources.

How I Did it
------------
Updated the `KustomizeOverlay` component to check for files that `=== isResource`, and save/delete accordingly if so.

How to verify it
------------
Run `./bin/ship init http://github.com/helm/master/stable/bitcoind --rm-asset-dest`
Make at least 1 patch
In the state file, copy/paste the patches object, and change the key to `resources`, also change the file name(s).
Run `./bin/ship update --headed --prefer-git`

Description for the Changelog
------------
Updated the ship init UI flow to handle new resources.

------------
basketball bois

### DIFF
--- a/web/init/src/redux/data/kustomizeOverlay/reducer.js
+++ b/web/init/src/redux/data/kustomizeOverlay/reducer.js
@@ -14,7 +14,7 @@ function updateFileContents(currState, data) {
     overlayContent: data.content.overlay,
     key: data.path,
     isSupported: data.content.isSupported,
-    isResource: data.content.isResource || null
+    isResource: data.content.isResource
   }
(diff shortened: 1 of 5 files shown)


web/init/src/redux/data/kustomizeOverlay/reducer.js
```diff
@@ -14,6 +14,7 @@ function updateFileContents(currState, data) {
     overlayContent: data.content.overlay,
     key: data.path,
     isSupported: data.content.isSupported,
+    isResource: data.content.isResource || null
```
Rob0h: is it necessary to be using `null` here?

PR #646, opened 2018-10-17
Allow ship to be pointed to git file paths

Rob0h:
What I Did
------------
- Allow ship to be pointed to git file paths

How I Did it
------------
- Add additional logic for how files are determined to be included from the tar

How to verify it
------------
- Tests
- `ship init github.com/scalyr/scalyr-agent-2/blob/master/k8s/scalyr-agent-2.yaml`

Description for the Changelog
------------
- Allow ship to be pointed to git file paths

------------
🛶

### DIFF
--- a/pkg/specs/githubclient/client.go
+++ b/pkg/specs/githubclient/client.go
@@ -145,7 +145,9 @@ func (g *GithubClient) downloadAndExtractFiles(
 				}
 				basePathFound = true
 
-				fileName = strings.TrimPrefix(fileName, basePath)
+				if fileName != basePath {
+					fileName = strings.TrimPrefix(fileName, basePath)
+				}
 				outFile, err := g.fs.Create(filepath.Join(filePath, fileName))
 				if err != nil {
 					return errors.Wrapf(err, "extract tar gz, create")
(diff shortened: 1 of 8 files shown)


laverya:
Could you add an integration test with a blob upstream?

pkg/specs/githubclient/client.go
```diff
@@ -165,6 +167,21 @@ func (g *GithubClient) downloadAndExtractFiles(
...
+
+	if includeIsFile {
+		if currentIsFile {
+			return currentPath == includePath
+		}
+
+		return strings.HasPrefix(includePath, currentPath)
+	}
+
+	return strings.HasPrefix(currentPath, includePath)
```
laverya: I feel like this will break on a couple of situations, the most obvious one being when you have a directory name that is a substring of another directory name:
currentPath `dir/file2/blah`, includePath `dir/file`

We might have written code for this elsewhere in Ship

laverya:
I'll add that in a followup PR, you don't have to deal with it here

pkg/specs/githubclient/client.go
laverya: We could also just not try and special case the "singleton file" bit here and instead always create the parent directory for files before creating the files
That would require the "mkdir -p" equivalent to exist, of course

PR #647, opened 2018-10-17
Remove watch dir after checking content sha

marccampbell:
What I Did
------------
When `ship watch` sleeps, it should start with a clean filesystem each time. Currently, this was failing on the second pass if it was a go-getter url because there was no .git directory.

How I Did it
------------
Delete the directory immediately after checking the content sha in the watch command.

How to verify it
------------
Run watch on a ship app with a go-getter url and let the interval expire twice. There should be no more error.

Description for the Changelog
------------
- Fix a bug that would cause `watch` to crash on the second loop.

------------

### DIFF
--- a/pkg/specs/interface.go
+++ b/pkg/specs/interface.go
@@ -5,7 +5,6 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"net/url"
-	"os"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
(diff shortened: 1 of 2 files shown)


pkg/specs/interface.go
```diff
@@ -88,6 +89,12 @@ func (r *Resolver) ReadContentSHAForWatch(ctx context.Context, upstream string)
 	}
 	debug.Log("event", "apptype.inspect", "type", appType, "localPath", localPath)
 
+	defer func() {
+		if err := os.RemoveAll(localPath); err != nil {
```
Rob0h:
```
r.FS.RemoveAll(localPath)
``` 
?

marccampbell: 👍

PR #650, opened 2018-10-22
Add field in helmValues step to change releaseName

Rob0h:
What I Did
------------
Add field in helmValues step to change releaseName

How I Did it
------------
Serialize releaseName at resolveRelease step
Add field for UI in helmValues step to change releaseName

How to verify it
------------
- Tests (for serialization of releaseName)
- Run `ship init` against any helm chart, Change the releaseName at the helmValues step, Observe changes to rendered yaml

Description for the Changelog
------------
Add field in helmValues step to change releaseName

------------
🛶

### DIFF
--- a/pkg/api/spec.go
+++ b/pkg/api/spec.go
@@ -1,5 +1,12 @@
 package api
 
+import (
+	"regexp"
+	"strings"
+)
+
+var releaseNameRegex = regexp.MustCompile("[^a-zA-Z0-9\\-]")
+
 // Spec is the top level Ship document that defines an application
 type Spec struct {
 	Assets    Assets    `json:"assets" yaml:"assets" hcl:"asset"`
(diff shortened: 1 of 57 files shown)


pkg/api/spec.go
```diff
@@ -1,5 +1,12 @@
 package api
 
+import (
+	"regexp"
+	"strings"
+)
+
+var releaseNameRegex = regexp.MustCompile("[^a-zA-Z0-9\\-]")
```
dexhorthy: Its not clear from the name of this what its for -- it looks like it matches characters that are disallowed in a release name?

pkg/lifecycle/daemon/routes_v1.go
```diff
@@ -122,12 +123,19 @@ func (d *V1Routes) saveHelmValues(c *gin.Context) {
...
-		debug.Log("event", "seralize.fail", "err", err)
+	debug.Log("event", "serialize.helmValues")
+	if err := d.StateManager.SerializeHelmValues(request.Values, string(chartDefaultValues)); err != nil {
+		debug.Log("event", "seralize.helmValues.fail", "err", err)
 		c.AbortWithError(http.StatusInternalServerError, errors.New("internal_server_error"))
 	}
+
+	debug.Log("event", "serialize.helmReleaseName")
+	if len(request.ReleaseName) > 0 {
+		if err := d.StateManager.SerializeReleaseName(request.ReleaseName); err != nil {
```
dexhorthy: why are we having the client set the release name?

pkg/api/spec.go
dexhorthy: I see that this is copy pasted, so I'm :shrug: on it.

pkg/lifecycle/render/helm/template_test.go
```diff
@@ -82,7 +83,7 @@ func TestLocalTemplater(t *testing.T) {
 			expectedHelmValues: []string{
```
dexhorthy: judging by the name of this test, I'd imagine its no longer required.

pkg/state/models.go
```diff
@@ -14,6 +14,7 @@ type State interface {
 	CurrentKustomizeOverlay(filename string) (string, bool)
 	CurrentHelmValues() string
 	CurrentHelmValuesDefaults() string
+	CurrentReleaseName() string
```
dexhorthy: I don't think we need to continue adding "CurrentXXXX" methods to the `State` interface -- you should be able to just use `.Versioned()` and then access the release name directly?

pkg/lifecycle/daemon/routes_v1.go
Rob0h: it's a client request 😅

PR #653, opened 2018-10-22
[ch15317] Added no-outro flag for optional confirm step

10Dimensional:
What I Did
------------
Added a `--no-outro` flag to optionally remove the outro step from the ship lifecycle.

How I Did it
------------
Added boolean `--no-outro` flag in `pkg/cli/root.go`. Then check for the flag when creating the ship lifecycle.

How to verify it
------------
Run `ship init` with `--no-outro`

Description for the Changelog
------------
Added a `--no-outro` flag to optionally remove the outro step from the ship lifecycle.

------------
⛵️

### DIFF
--- a/pkg/specs/chart.go
+++ b/pkg/specs/chart.go
@@ -41,17 +41,9 @@ func (r *Resolver) DefaultHelmRelease(chartPath string) api.Spec {
 		Lifecycle: api.Lifecycle{
 			V1: []api.Step{
 				{
-					HelmIntro: &api.HelmIntro{
-						StepShared: api.StepShared{
-							ID: "intro",
-						},
-					},
-				},
-				{
 					HelmValues: &api.HelmValues{
 						StepShared: api.StepShared{
 							ID:          "values",
-							Requires:    []string{"intro"},
 							Invalidates: []string{"render"},
 						},
 					},
--- a/pkg/specs/chart.go
+++ b/pkg/specs/chart.go
@@ -19,8 +19,9 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-func DefaultHelmRelease(chartPath string) api.Spec {
-	return api.Spec{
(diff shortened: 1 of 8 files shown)


pkg/specs/chart.go
```diff
@@ -19,8 +19,9 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-func DefaultHelmRelease(chartPath string) api.Spec {
-	return api.Spec{
+func (r *Resolver) DefaultHelmRelease(chartPath string) api.Spec {
+	noOutro := r.Viper.GetBool("no-outro")
```
dexhorthy: I think in general we prefer to set these as fields on `(r *Resolver)` where possible in the `New` function, rather than accessing `r.Viper` inside a method. 

For example, check out https://github.com/replicatedhq/ship/blob/master/pkg/ship/ship.go#L79

laverya: Yeah, that's my fault... still probably the better pattern to follow though @10Dimensional

pkg/specs/constructor.go
```diff
@@ -53,5 +54,6 @@ func NewResolver(
 		},
 		AppResolver:               appresolver,
 		GitHubReleaseNotesFetcher: github,
+		NoOutro:                   v.GetBool("no-outro"),
```
dexhorthy: nice :+1:

pkg/specs/chart.go
```diff
@@ -40,13 +40,6 @@ func (r *Resolver) DefaultHelmRelease(chartPath string) api.Spec {
...
 			V1: []api.Step{
-				{
-					HelmIntro: &api.HelmIntro{
-						StepShared: api.StepShared{
-							ID: "intro",
-						},
-					},
-				},
 				{
 					HelmValues: &api.HelmValues{
```
dexhorthy: I think you need to remove the dependency on `intro` in this step

PR #658, opened 2018-10-23
fix nil pointer panic

dexhorthy:
What I Did
------------

Fix a nil pointer panic when walking local FS with `set-github-contents`

How I Did it
------------

Don't call `info.Name()`, it might be nil, just use `info`, and hope `path` has what we need.

How to verify it
------------

--set-github-contents with a non-existent base path

Description for the Changelog
------------

Fix a nil pointer panic when walking local FS with `set-github-contents`

:boat:

### DIFF
--- a/pkg/specs/replicatedapp/local.go
+++ b/pkg/specs/replicatedapp/local.go
@@ -81,7 +81,7 @@ func (r *resolver) loadLocalGithubFiles(localpath string, repoPath string) ([]Gi
 	var files []GithubFile
 	err := r.FS.Walk(localpath, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
-			return errors.Wrapf(err, "walk %s from %s", info, path)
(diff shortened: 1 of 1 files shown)


pkg/specs/replicatedapp/local.go
```diff
@@ -81,7 +81,7 @@ func (r *resolver) loadLocalGithubFiles(localpath string, repoPath string) ([]Gi
 	var files []GithubFile
 	err := r.FS.Walk(localpath, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
-			return errors.Wrapf(err, "walk %s from %s", info.Name(), path)
+			return errors.Wrapf(err, "walk %s from %s", info, path)
```
laverya:
```suggestion
			return errors.Wrapf(err, "walk %+v from %s", info, path)
```
I don't think info is a string, is it?

PR #660, opened 2018-10-23
Grab bag of small usability fixes

dexhorthy:
What I Did
------------

- flag to apply terraform in headless mode
- dont map falsey items in UI
  - I dont know why some of these are making it into props as "undefined", but
    this fixes for now
- use render root in terraform steps

How to verify it
------------

- Try `--terraform-apply-yes` in conjunction with `--headless` flag
- Run a ship.yaml with terraform **and** a non-default lifecycle.v1.render.root

Description for the Changelog
------------

- Fix a bug in terraform where render.root was not respected
- Add flag `--terraform-apply-yes` for usage with ship in `--headless` mode.

### DIFF
--- a/pkg/lifecycle/terraform/daemonless.go
+++ b/pkg/lifecycle/terraform/daemonless.go
@@ -13,10 +13,10 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/ship/pkg/api"
-	"github.com/replicatedhq/ship/pkg/constants"
 	"github.com/replicatedhq/ship/pkg/lifecycle"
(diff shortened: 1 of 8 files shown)


pkg/lifecycle/terraform/daemonless.go
```diff
@@ -55,19 +57,24 @@ func (t *DaemonlessTerraformer) WithStatusReceiver(
...
 		Terraform:     t.Terraform,
 		Viper:         t.Viper,
+		FS:            t.FS,
 
 		Status: statusReceiver,
 	}
 }
 
 func (t *DaemonlessTerraformer) Execute(ctx context.Context, release api.Release, step api.Terraform, confirmedChan chan bool) error {
-	t.dir = step.Path
```
dexhorthy: kind of tangential -- I want to keep runtime state on the stack, not in struct fields if possible

PR #661, opened 2018-10-24
store terraform state

dexhorthy:
What I Did
------------

Store terraform state after terraform apply runs. Restore it before "init".

How I Did it
------------

- new function `persistState` used by `Daemonless` and `Fork` terraformers.
(spolier alert, they both fork)
- vendor in subset of terraform code. Vendoring in the whole thing is still
tricky, but I was able to get the types loaded in, which lets us use that
library for unmarshalling the state.
- store state after "apply" runs

How to verify it
------------

Run a ship spec with terraform in it, inspect state.json afterward

Description for the Changelog
------------
Store state from `terraform` in ship state.json

:ship:


integration/base/terraform/expected/.ship/release.yml
```diff
@@ -3,13 +3,15 @@ assets:
     - terraform:
         dest: ./terraform/random.tf
         inline: |
-          resource "random_id" "server" {
-            byte_length = {{repl ConfigOption "id_length" | ParseInt}}
+          resource "local_file" "foo" {
+            content     = "foo!"
+            filename = "/tmp/foo.bar"
```
laverya: do we no longer support template functions here?

pkg/lifecycle/terraform/daemonless.go
```diff
@@ -58,18 +68,26 @@ func (t *DaemonlessTerraformer) WithStatusReceiver(
 		Terraform:     t.Terraform,
 		Viper:         t.Viper,
 		FS:            t.FS,
+		StateManager:  t.StateManager,
+		StateSaver:    t.StateSaver,
```
dexhorthy: need a restorer here

pkg/lifecycle/terraform/daemonless.go
```diff
@@ -161,7 +185,7 @@ func (t *DaemonlessTerraformer) plan(dir string) (string, bool, error) {
 
 	// we really shouldn't write plan to a file, but this will do for now
 	cmd := t.Terraform(dir)
-	cmd.Args = append(cmd.Args, "plan", "-input=false", "-out=plan")
+	cmd.Args = append(cmd.Args, "plan", "-input=false", "-out=plan.tfplan")
```
dexhorthy: added this extension

pkg/lifecycle/terraform/terraformer.go
```diff
@@ -64,19 +72,28 @@ func (t *ForkTerraformer) WithStatusReceiver(status daemontypes.StatusReceiver)
 		Terraform:     t.Terraform,
 		Viper:         t.Viper,
 		FS:            t.FS,
+		StateManager:  t.StateManager,
```
dexhorthy: this needs persistState/restoreState

pkg/lifecycle/terraform/terraformer.go
```diff
@@ -135,7 +155,12 @@ func (t *ForkTerraformer) Execute(ctx context.Context, release api.Release, step
 		<-t.Daemon.MessageConfirmedChan()
 	}
 
-	return nil
+	err = t.StateSaver(debug, t.FS, t.StateManager, dir)
+	if err != nil {
+		return errors.Wrapf(err, "persist terraform state")
+	}
+
+	return err
```
dexhorthy: this should be nil

laverya: Either make this a `return nil` or `return errors.Wrapf(err, "persist terraform state")` and get rid of the conditional

```suggestion
	return errors.Wrapf(err, "persist terraform state")
```

`errors.Wrap(nil, "string) == nil`

integration/base/terraform/expected/.ship/release.yml
dexhorthy: we do, I was just trying to make these more reproducible. I can actually roll some of that back now that we're restoring the state

PR #665, opened 2018-10-25
Ship Init file::./ship.yaml

dexhorthy:
What I Did
------------

Support `ship init`-ing a file, directory or github URL that contains just a
ship YAML. We'll read the release from the `ship.yaml` instead of a
default release. We'll also skip the "copy the upstream into `./base`"
step for these apps.

How I Did it
------------

- Add new app type `inline.replicated.app`
- Check for `ship.yaml` in `apptype.Inspector`
- Add switch blocks for `inline.replicated.app`

How to verify it
------------

There's an integration test, but also you can now

```
ship init github.com/replicatedhq/test-charts/tree/master/just-ship-yaml
```

Description for the Changelog
------------

Add support for ship init file::./ship.yaml

:ship:

### DIFF
--- a/integration/init/just-ship-yaml/metadata.yaml
+++ b/integration/init/just-ship-yaml/metadata.yaml
@@ -1,0 +1,3 @@
+upstream: "https://github.com/replicatedhq/test-charts/tree/ebebc9e692db3caeadb308ddeec37e9565acba1e/just-ship-yaml"
+args: []
+skip_cleanup: true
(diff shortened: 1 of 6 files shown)


integration/init/just-ship-yaml/metadata.yaml
```diff
@@ -0,0 +1,3 @@
+upstream: "https://github.com/replicatedhq/test-charts/tree/ebebc9e692db3caeadb308ddeec37e9565acba1e/just-ship-yaml"
+args: []
```
laverya:
```suggestion
args: ["--prefer-git"]
```
Since that doesn't hit GitHub rate limits which makes it easier to run the integration tests locally

PR #672, opened 2018-10-30
Support k8s list for init

Rob0h:
What I Did
------------
- Support k8s list for init

How I Did it
------------
- Split list yaml
- Store information on split into state.json
- Reassemble list yaml based on state.json

How to verify it
------------
- Integration
- run `ship init https://raw.githubusercontent.com/jaegertracing/jaeger-kubernetes/master/jaeger-production-template.yml`

Description for the Changelog
------------
Support k8s list for init

------------
🛶


pkg/specs/chart.go
```diff
@@ -390,3 +376,91 @@ func (r *Resolver) maybeSplitMultidocYaml(ctx context.Context, localPath string)
...
+	if err != nil {
+		return errors.Wrapf(err, "read files in %s", path)
+	}
+
+	var lists []state.List
+	for _, file := range files {
+		filePath := filepath.Join(path, file.Name())
+
+		if file.IsDir() {
+			// TODO: handling nested list yamls
```
Rob0h: not sure of how important this is / how often it will come up

integration/init/jaeger/metadata.yaml
```diff
@@ -0,0 +1,3 @@
+upstream: "https://github.com/jaegertracing/jaeger-kubernetes/blob/59e7afcae7b2f3bc109695cfda2a5c116cea4391/jaeger-production-template.yml"
+args: []
```
laverya:
```suggestion
args: ["--prefer-git"]
```
To help the GitHub API rate limits of people running the integration tests locally

pkg/specs/chart.go
laverya: Perhaps call the function recursively and then `continue`? Either way, I know you need to add a `continue` here or things will blow up when it hits the ReadFile call below

pkg/specs/chart.go
```diff
@@ -390,3 +376,91 @@ func (r *Resolver) maybeSplitMultidocYaml(ctx context.Context, localPath string)
...
 	return nil
 }
+
+type ListK8sYaml struct {
+	APIVersion string        `json:"apiVersion" yaml:"apiVersion"`
+	Kind       string        `json:"kind" yaml:"kind" hcl:"kind"`
+	Items      []interface{} `json:"items" yaml:"items"`
+}
+
+func (r *Resolver) maybeSplitListYaml(ctx context.Context, path string) error {
```
laverya: Could we get a test for this function? You should be able to largely copy https://github.com/replicatedhq/ship/blob/master/pkg/specs/chart_test.go#L125 for the test code

laverya:
@Rob0h this look like a good idea for an integration test? https://github.com/replicatedhq/test-charts/pull/8

PR #675, opened 2018-11-01
Updated the Helm and Kustomize intro step images and content

10Dimensional:
What I Did
------------
Updated the Helm and Kustomize intro step images and content. Also added `isUpdate` key on `HelmIntro` type to allow for different copy on the intro page depending on whether a person is going through the `init` or `update` flow.

How I Did it
------------
- Replaced old copy and images on intro steps
- Set an `isUpdate` flag to true on the update command, and get it in the helm intro step

How to verify it
------------
Run ship init or ship update and on a helm chart

Description for the Changelog
------------
Updated the Helm and Kustomize intro step images and content.

------------
⛵️


pkg/lifecycle/helmIntro/helmIntro.go
```diff
@@ -17,6 +17,7 @@ import (
 )
 
 type HelmIntro struct {
+	v      *viper.Viper
```
laverya: I don't think you ever actually use this viper

PR #676, opened 2018-11-01
add build-minimal target and env options to enable git-free builds for homebrew

fshot:
What I Did
------------
* Added a `build-minimal` make target (which builds the complete go binaries & skips some git-related hygiene steps)
* Added env-variable options for specifying VERSION and GIT_SHA when invoking make, so the homebrew build-from-source will work fine in a tarball that's not a git repo

How I Did it
------------

How to verify it
------------
make VERSION="0.21.0.testing" GIT_SHA="" build-minimal
make VERSION="0.21.0.testing" build-minimal
make build-minimal
etc...

Description for the Changelog
------------
add build-minimal target and env options to enable git-free builds for homebrew

------------
concreteboatplans

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -261,9 +261,9 @@ build: fmt embed-ui-dev test bin/ship
 
 build-ci: ci-embed-ui bin/ship
 
-build-ci-cypress: mark-ui-gitignored pkg/lifeycle/daemon/ui.bindatafs.go bin/ship
+build-ci-cypress: mark-ui-gitignored pkg/lifecycle/daemon/ui.bindatafs.go bin/ship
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -254,6 +263,8 @@ build-ci: ci-embed-ui bin/ship
 
 build-ci-cypress: mark-ui-gitignored pkg/lifeycle/daemon/ui.bindatafs.go bin/ship
 
+build-minimal: bin/ship build-ui pkg/lifeycle/daemon/ui.bindatafs.go
```
laverya:
```suggestion
build-minimal: build-ui pkg/lifecycle/daemon/ui.bindatafs.go bin/ship
```
As suggested by @dexhorthy 
Also, lifecycle was misspelled (and it is in `build-ci-cypress` too...)

PR #682, opened 2018-11-06
Allow ship init to be pointed to valid go-getter ship.yaml path

Rob0h:
What I Did
------------
- Allow ship init to be pointed to valid go-getter ship.yaml path

How I Did it
------------
- Point ship init to a ship.yaml

How to verify it
------------
Point `ship init` to valid go-getter ship.yaml path

Description for the Changelog
------------
Allow ship init to be pointed to valid go-getter ship.yaml path

------------
🛶

### DIFF
--- a/integration/init/integration_test.go
+++ b/integration/init/integration_test.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"testing"
 
 	"github.com/docker/docker/client"
@@ -19,9 +20,9 @@ import (
 )
 
 type TestMetadata struct {
-	Upstream string   `yaml:"upstream"`
-	Args     []string `yaml:"args"`
-
+	Upstream     string   `yaml:"upstream"`
+	Args         []string `yaml:"args"`
+	MakeAbsolute bool     `yaml:"make_absolute"`
 	// debugging
 	SkipCleanup bool `yaml:"skip_cleanup"`
 }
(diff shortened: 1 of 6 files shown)


integration/init/local-single-file-gogetter/expected/.ship/state.json
```diff
@@ -0,0 +1,9 @@
+{
+  "v1": {
+    "config": null,
+    "releaseName": "ship",
+    "upstream": "file::/Users/robert/go/src/github.com/replicatedhq/ship/integration/init/local-single-file-gogetter/input/ship.yaml",
```
laverya: Not OK

Maybe change this to `<UPSTREAM>` and include the upstream as part of the string replacement? (the map[string]string{} on line 94 of integration/init/integration_test.go)

pkg/specs/apptype/determine_type.go
```diff
@@ -137,10 +140,9 @@ func (r *inspector) determineTypeFromContents(
 		if err != nil {
 			return "", "", errors.Wrapf(err, "check for %s", filename)
 		}
-	}
-
-	if isReplicatedApp {
-		return "inline.replicated.app", finalPath, nil
+		if isReplicatedApp {
```
Rob0h: moved this up because it checks for 2 different files, if the first is true and the second is false... 😢

integration/init/local-single-file-gogetter/expected/.ship/state.json
Rob0h: oh yea this is why i didn't push this

laverya: We do variable substitution for all the `init_app` tests: https://github.com/replicatedhq/ship/blob/master/integration/init_app/amazon-eks-template/expected/.ship/state.json
Seems like the idiom is `__upstream__` not `<UPSTREAM>` though

integration/init/integration_test.go
```diff
@@ -78,20 +79,33 @@ var _ = Describe("ship init with arbitrary upstream", func() {
...
+
+					if testMetadata.MakeAbsolute {
+						relativePath := testMetadata.Upstream
+						pwdRoot, err := os.Getwd()
+						Expect(err).NotTo(HaveOccurred())
+						pwdRoot, err = filepath.Abs(pwdRoot)
+						Expect(err).NotTo(HaveOccurred())
+						absolutePath := filepath.Join(pwdRoot, "..")
+						absoluteUpstream = fmt.Sprintf("file::%s", filepath.Join(absolutePath, relativePath))
+						replacements["__upstream__"] = absolutePath
```
laverya: I would prefer 
```suggestion
						replacements["__upstream__"] = absoluteUpstream
```

integration/init/local-single-file-gogetter/expected/.ship/state.json
```diff
@@ -0,0 +1,9 @@
+{
+  "v1": {
+    "config": null,
+    "releaseName": "ship",
+    "upstream": "file::__upstream__/input/ship.yaml",
```
laverya: and
```suggestion
    "upstream": "file::__upstream__",
```

PR #686, opened 2018-11-07
Fix for Istio 1.0.3

Rob0h:
What I Did
------------
Fix for Istio 1.0.3

How I Did it
------------
Only update helm dependencies if the repo is defined

How to verify it
------------
Integration test for Istio 1.0.3

Description for the Changelog
------------
Fix for Istio 1.0.3

------------
🛶


pkg/lifecycle/render/helm/commands.go
```diff
@@ -37,16 +46,45 @@ func (h *helmCommands) Init() error {
...
+		return errors.Wrap(err, "read requirements yaml")
 	}
-	return dependencyCommand.Execute()
+
+	requirements := chartutil.Requirements{}
+	if err := yaml.Unmarshal(requirementsB, &requirements); err != nil {
+		return errors.Wrap(err, "unmarshal requirements yaml")
+	}
+
+	allEmpty := true
```
Rob0h: the only case i really want to deal with right now is where all the repository fields are empty - e.g. istio

integration/init/istio-1.0.3/metadata.yaml
```diff
@@ -0,0 +1,3 @@
+upstream: "https://github.com/istio/istio/tree/1.0.3/install/kubernetes/helm/istio"
+args: []
```
laverya:
```suggestion
args: ["--prefer-git"]
```
We should include this whenever possible

PR #695, opened 2018-11-12
Delete base resources from UI

Rob0h:
What I Did
------------
- Delete base resources from UI

How I Did it
------------
- Add an `excludedBases` field to `state.json` which is used when writing the base `kustomization.yaml`

How to verify it
------------
- Tests, Integration
- Run `ship init` and delete a base from the ui at the kustomize step

Description for the Changelog
------------
- Delete base resources from UI

------------
🛶

### DIFF
--- a/integration/update/excluded-basic/metadata.yaml
+++ b/integration/update/excluded-basic/metadata.yaml
@@ -1,0 +1,2 @@
+args: []
+skip_cleanup: false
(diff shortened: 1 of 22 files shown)


integration/update/excluded-basic/metadata.yaml
```diff
@@ -0,0 +1,2 @@
+args: []
```
laverya:
```suggestion
args: ["--prefer-git"]
```

PR #704, opened 2018-11-14
Download local/no proxy github assets through ship

Rob0h:
What I Did
------------
- Download local/no proxy github assets through ship

How I Did it
------------
- During the render phase, if no matching files are found for the github asset and the asset is local or has proxy defined as false - ship will download the assets to the local fs

How to verify it
------------
- Integration tests
- Run `ship app` pointed to a runbook with github assets defined

Description for the Changelog
------------
- Download local/no proxy github assets through ship

------------
🛶


pkg/lifecycle/render/github/render.go
```diff
@@ -82,50 +95,110 @@ func (r *LocalRenderer) Execute(
...
+	return nil
+}
+
+func (r *LocalRenderer) resolveNoProxyGithubAssets(asset api.GitHubAsset, builder *templates.Builder) error {
+	debug := level.Debug(log.With(r.Logger, "step.type", "render", "render.phase", "execute", "asset.type", "github", "dest", asset.Dest, "description", asset.Description))
+	debug.Log("event", "createUpstream")
+	upstream := createUpstreamURL(asset)
+
+	var fetcher apptype.FileFetcher
+	localFetchPath := filepath.Join(constants.InstallerPrefixPath, constants.GithubAssetSavePath)
```
Rob0h: `rootFs` does not work well with `gogetter`

PR #706, opened 2018-11-15
Maintain helm values keys order

emosbaugh:
What I Did
------------
Ship init Helm values now appear in order after saving and navigating back to the /values page.

How I Did it
------------
Merging helm values from the upstream and the ship state now uses yaml.MapSlice to maintain order of keys.

How to verify it
------------
Save the /values page and then navigate back. Keys should appear in order.

Description for the Changelog
------------
Ship init helm value keys will now maintain order after saving and returning to the /values page in the nav cycle.

------------


pkg/lifecycle/render/helm/values_test.go
```diff
@@ -4,17 +4,26 @@ import (
...
+		m2 := yaml.MapSlice{
+			{Key: "b", Value: true},
+		}
+		m3 := yaml.MapSlice{
+			{Key: "a", Value: "value"},
+			{Key: "b", Value: false},
+			{Key: "c", Value: nil},
+		}
 		allKeys := getAllKeys(m1, m2, m3)
 		req.Contains(allKeys, "a")
```
laverya: should we assert that the ordering is preserved here?

PR #712, opened 2018-11-17
Move list yaml splitting to render

Rob0h:
What I Did
------------
- Move list yaml splitting to render

How I Did it
------------
- Move list yaml splitting to render step, adds render step as part of the default lifecycle steps in k8s yaml

How to verify it
------------
- Integration for jaeger and cassandra, included as github assets

Description for the Changelog
------------
- Move list yaml splitting to render

------------
🛶


pkg/lifecycle/daemon/daemon_test.go
```diff
@@ -194,6 +196,7 @@ func TestDaemonAPI(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
+			fmt.Println("name", test.name)
```
jadengore: Kill this

pkg/lifecycle/daemon/routes_navcycle_getstep.go
```diff
@@ -26,6 +27,16 @@ func (d *NavcycleRoutes) getStep(c *gin.Context) {
 				return
 			}
 
+			if preExecuteFunc, exists := d.PreExecuteFuncMap[step.ShortName()]; exists {
+				if err := preExecuteFunc(context.Background(), step); err != nil {
```
jadengore: Should we be using background context here? What do these preExecuteFuncs do?

jadengore: OK, looks like your current use of this does not use context yet, but this is still worth considering before we merge.

Rob0h: discussed irl - background context from the request handler is passed to pre execute funcs

PR #719, opened 2018-11-21
ship watch prints No changes detected

areed:
What I Did
------------
Add output to ship watch when no changes are detected.

How I Did it
------------
UI.Info

How to verify it
------------
Run `ship watch`

Description for the Changelog
------------
Print "No changes detected" in ship watch after checking upstream.

------------

### DIFF
--- a/pkg/ship/watch.go
+++ b/pkg/ship/watch.go
@@ -75,9 +75,10 @@ func (s *Ship) Watch(ctx context.Context) error {
 			"sleeping", s.Viper.GetDuration("interval"),
 		)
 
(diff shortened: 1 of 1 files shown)


pkg/ship/watch.go
```diff
@@ -75,6 +75,7 @@ func (s *Ship) Watch(ctx context.Context) error {
 			s.UI.Info(noUpdateMsg)
 			return nil
 		}
+		s.UI.Info("No changes detected")
```
laverya: Would it be a good idea to combine this with the noUpdateMsg above?

PR #721, opened 2018-11-21
Allow istio to be updated based on version releases

Rob0h:
What I Did
------------
- Allow istio to be updated based on version releases

How I Did it
------------
- If the upstream is a valid git url, detect a version and replace that with `_latest_` token
- The `_latest_` token is replaced during `ship watch` and `ship update` with the latest release found on github

How to verify it
------------
- Run `ship init` pointed to an older version of istio
- Run `ship update` to be updated to the latest version

Description for the Changelog
------------
- Allow istio to be updated based on version releases

------------
🛶


integration/init/istio-1.0.3/metadata.yaml
```diff
@@ -1,3 +1,3 @@
 upstream: "https://github.com/istio/istio/tree/1.0.3/install/kubernetes/helm/istio"
 args: ["--prefer-git"]
-skip_cleanup: false
+skip_cleanup: true
```
laverya:
```suggestion
skip_cleanup: false
```

pkg/specs/upstream_test.go
```diff
@@ -0,0 +1,185 @@
...
+			expected:  "",
+			expectErr: true,
+		},
+		{
+			name:     "not a versioned upstream",
+			upstream: "github.com/o/r/tree/ref",
+			currentState: &state.VersionedState{
+				V1: &state.V1{
+					Metadata: &state.Metadata{
+						Version: "1.2.0",
```
laverya: what happens with a ref upstream when no version is in the state?

laverya: Also, should we test upstreams like "github.com/o/r/tree/ref/path" and "notgithub.com/o/r/tree/1.0.1"?

pkg/specs/upstream_test.go
```diff
@@ -0,0 +1,185 @@
...
+		expected string
+	}{
+		{
+			name:     "no version - no change",
+			upstream: "github.com/helm/charts/stable/grafana",
+			expected: "github.com/helm/charts/stable/grafana",
+		},
+		{
+			name:     "has a version - changed",
+			upstream: "https://github.com/istio/istio/tree/1.0.2/install/kubernetes/helm/istio",
```
laverya: can we add a 'ref' version of this? `https://github.com/istio/istio/tree/abcdef/install/kubernetes/helm/istio` or something similar

pkg/specs/upstream_test.go
Rob0h: 1. ref upstream remains unchanged
2. added all the test cases

pkg/specs/upstream.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+
+		maybeVersionedUpstream = strings.Replace(upstream, UpstreamVersionToken, latestReleaseVersion, 1)
+
+		debug.Log("event", "check previous version")
+		if existingState.Versioned().V1.Metadata != nil && existingState.Versioned().V1.Metadata.Version != "" {
+			if strings.Contains(upstream, UpstreamVersionToken) {
+				previousVersion, err := version.NewVersion(existingState.Versioned().V1.Metadata.Version)
+				if err != nil {
+					return maybeVersionedUpstream, nil
```
laverya: why are we returning `nil` here and in the next block? Are these errors expected?

pkg/specs/upstream.go
```diff
@@ -0,0 +1,76 @@
...
+				}
+
+				latestVersion, err := version.NewVersion(latestReleaseVersion)
```
laverya: `latestVersion` is never returned

Rob0h: no the intent is to return `maybeVersionedUpstream`, `latestVersion` is only a `go-version` version used for comparison purposes

pkg/specs/upstream.go
Rob0h: for whatever reason if the previous version parsing returns an error it will just fall back to returning the `maybeVersionedUpstream`

PR #725, opened 2018-11-27
 support for ship init path/to/local/ship.yaml

emosbaugh:
What I Did
------------
There were two different code paths that have similar functionality, one `ship app` and the other `ship init path/to/local/ship.yaml`. I've merge all additional functionality from `ship app` into `ship init` and removed `ship app` command.

How I Did it
------------
Changed the code.

How to verify it
------------
1) Run the integration tests
2) https://github.com/retracedhq/api/pull/389

Description for the Changelog
------------
The `ship app --runbook /path/to/local/ship.yaml` command has been remove in favor of `ship init /path/to/local/ship.yaml`

------------


integration/base/integration_test.go
```diff
@@ -141,7 +143,9 @@ var _ = Describe("ship app", func() {
 					Expect(err).NotTo(HaveOccurred())
 
 					//compare the files in the temporary directory with those in the "expected" directory
-					result, err := integration.CompareDir(path.Join(testPath, "expected"), testOutputPath, map[string]string{})
+					result, err := integration.CompareDir(path.Join(testPath, "expected"), testOutputPath, map[string]string{
+						"UPSTREAM": upstream,
```
laverya: Could we use `__upstream__` here, to match the other test suites?

pkg/cli/init.go
```diff
@@ -50,9 +58,17 @@ Upstream can be one of:
 	cmd.Flags().Bool("rm-asset-dest", false, "Always remove asset destinations if already present")
 	cmd.Flags().Int("retries", 3, "Number of times to retry retrieving upstream")
 
-	viper.BindPFlags(cmd.Flags())
-	viper.BindPFlags(cmd.PersistentFlags())
-	viper.AutomaticEnv()
-	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
+	// optional flags for "ship init replicated.app"
+	cmd.Flags().String("customer-id", "", "Customer ID for which to query app specs. Required for all ship operations.")
```
laverya: I don't think this is `Required for all ship operations` anymore

integration/base/integration_test.go
```diff
@@ -87,16 +87,16 @@ var _ = Describe("ship app", func() {
 					buf := new(bytes.Buffer)
 					cmd.SetOutput(buf)
 					args := []string{
-						"app",
-						"--headless",
-						fmt.Sprintf("--runbook=%s", path.Join(testInputPath, ".ship/release.yml")),
+						"init", "staging.replicated.app/integration",
+						fmt.Sprintf("--runbook=%s", path.Join(testInputPath, ".ship/ship.yml")),
 						fmt.Sprintf("--customer-id=%s", testMetadata.CustomerID),
 						fmt.Sprintf("--installation-id=%s", testMetadata.InstallationID),
```
dexhorthy: Couple things
### online tests
 I think  for online tests, `customer-id` and `installation-id` should be expected to be in  the querystring, not passed as flags. e.g. `"staging.replicated.app/integration?customer_id=xxxx&installation_id=xxxxx"`

I don't think we should support these `--customer-id` and `--installation-id` flags for `ship init`, but leaving them in is fine for now. 

### Offline tests

For offline tests, we shouldn't use the `--runbook` flag, we should `ship init` on the runbook file `path.Join(testInputPath, ".ship/ship.yml")`

pkg/cli/init.go
laverya: Do we even want to support the `ship init replicated.app/<name> --customer-id=A --installation-id=B` form rather than the `replicated.app/<name>?customer_id=A&installation_id=B` form?

pkg/cli/init.go
```diff
@@ -50,9 +58,17 @@ Upstream can be one of:
...
+	cmd.Flags().String("installation-id", "", "Installation ID for which to query app specs")
```
dexhorthy: I think these should be removed in favor of querystring params

pkg/cli/init.go
```diff
@@ -50,9 +58,17 @@ Upstream can be one of:
...
+	cmd.Flags().String("release-semver", "", "Specific release version to pin installation to. Requires channel-id")
```
dexhorthy: maybe even ditto this one, a pinned semver is part of the `upstream`

pkg/cli/init.go
```diff
@@ -50,9 +58,17 @@ Upstream can be one of:
...
+
+	// optional developer flags for "ship init replicated.app"
+	cmd.Flags().String("runbook", "", developerFlagUsage)
```
dexhorthy: this one should go away, the `runbook` flag isn't necessary since we specify the `ship.yaml` to use via the argument to `init`

pkg/cli/init.go
```diff
@@ -50,9 +58,17 @@ Upstream can be one of:
...
+	cmd.Flags().String("set-channel-name", "", developerFlagUsage)
+	cmd.Flags().String("set-channel-icon", "", developerFlagUsage)
+	cmd.Flags().StringSlice("set-github-contents", []string{}, fmt.Sprintf("Specify a REPO:REPO_PATH:REF:LOCAL_PATH to override github checkouts to use a local path on the filesystem. %s. ", developerFlagUsage))
+	cmd.Flags().String("set-entitlements-json", "{\"values\":[]}", fmt.Sprintf("Specify json for entitlements payload. %s", developerFlagUsage))
```
dexhorthy: all the rest of these look good

pkg/specs/constructor.go
```diff
@@ -55,5 +60,10 @@ func NewResolver(
 		AppResolver:   appresolver,
 		GitHubFetcher: github,
 		NoOutro:       v.GetBool("no-outro"),
+
+		Runbook:        v.GetString("runbook"),
```
dexhorthy: I think this goes away too

pkg/specs/interface.go
```diff
@@ -75,13 +75,25 @@ func (r *Resolver) ResolveRelease(ctx context.Context, upstream string) (*api.Re
 			false,
 		)
 
+	case "runbook.replicated.app":
```
dexhorthy: what is this?

pkg/lifecycle/render/github/render.go
```diff
@@ -90,7 +90,7 @@ func (r *LocalRenderer) Execute(
 		debug.Log("event", "resolveProxyGithubAssets")
 		var files []api.GithubFile
 		for _, c := range meta.GithubContents {
-			if c.Repo == asset.Repo && c.Path == asset.Path && c.Ref == asset.Ref {
+			if c.Repo == asset.Repo && strings.Trim(c.Path, "/") == strings.Trim(asset.Path, "/") && c.Ref == asset.Ref {
```
dexhorthy: are there any tests around this change

integration/base/integration_test.go
dexhorthy: I guess if --customer-id is for local testing of template functions like `{{repl Installation "customer_id"}}` then it should stay, although potentially with a different flag name. the other use `--set-xxx` which seems to imply more that we're faking something for local dev. 

need to think about this one some more

pkg/specs/interface.go
emosbaugh: i need to return something different for that switch statement. i'm just following the pattern

pkg/lifecycle/render/github/render.go
emosbaugh: will add

PR #727, opened 2018-11-27
Run kustomize with base resources

Rob0h:
What I Did
------------
- Run kustomize with base resources

How I Did it
------------
- Run kustomize with base resources
- Move splitting of multidoc yaml to kustomize step and also deal with base resources recursively
- CRDs are left in their original state

How to verify it
------------
- Tests, Integration
- Run `ship init` and check the diff in the UI

Description for the Changelog
------------
- Run kustomize with base resources

------------
🚢


integration/init/cockroach/expected/base/Service-cockroachdb-public.yaml
```diff
@@ -1,21 +1,16 @@
...
-  # The secondary port serves the UI as well as health and debug endpoints.
-  - port: 8080
+  - name: http
+    port: 8080
     targetPort: 8080
-    name: http
   selector:
-    app: cockroachdb
\ No newline at end of file
+    app: cockroachdb
```
laverya: I don't really like how we're losing comments here
I suppose there's nothing for it though
Unless it's possible to use Ethan's comment preservation code... But I really highly doubt it

Rob0h: it might actually - i'll give it a go

pkg/lifecycle/kustomize/pre_kustomize.go
```diff
@@ -95,3 +109,87 @@ func (l *Kustomizer) maybeSplitListYaml(ctx context.Context, path string) error
...
+	}
+
+	if err := l.replaceOriginal(step, built); err != nil {
+		return errors.Wrap(err, "replace original yaml")
+	}
+
+	return nil
+}
+
+func (l *Kustomizer) replaceOriginal(step api.Kustomize, built []postKustomizeFile) error {
```
laverya: I would *love* a test for this
It looks like it should work fine with our mocked FS for testing, right?

laverya: Which exists already, and I completely missed. Right.

laverya:
(Also, none of the integration tests have a rendered.yaml change, so that's good)

PR #733, opened 2018-11-29
Fix step invalidation

Rob0h:
What I Did
------------
- Fix step invalidation

How I Did it
------------
- Allow steps to be executed multiple times in order for steps to be invalidated properly

How to verify it
------------
- Tests
- Run `ship init` and then execute a step that invalidates others, invalidated steps will execute again to show new results

Description for the Changelog
------------
- Fix step invalidation

------------
🛶

### DIFF
--- a/pkg/lifecycle/daemon/routes_navcycle_completestep.go
+++ b/pkg/lifecycle/daemon/routes_navcycle_completestep.go
@@ -12,7 +12,6 @@ import (
 	"github.com/replicatedhq/ship/pkg/api"
 	"github.com/replicatedhq/ship/pkg/lifecycle/daemon/daemontypes"
 	"github.com/replicatedhq/ship/pkg/lifecycle/daemon/statusonly"
-	"github.com/replicatedhq/ship/pkg/state"
 	"github.com/replicatedhq/ship/pkg/util/warnings"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/lifecycle/daemon/routes_navcycle_completestep.go
```diff
@@ -37,43 +36,25 @@ func (d *NavcycleRoutes) completeStep(c *gin.Context) {
...
-			}
-		}
-		_, stepAlreadyComplete := lifecycle.StepsCompleted[step.Shared().ID]
-
 		progress, ok := d.StepProgress.Load(step.Shared().ID)
-		shouldExecute := !ok || progress.Detail == `{"status":"success"}` && !stepAlreadyComplete
-
-		debug.Log("shouldExecute", shouldExecute)
+		shouldExecute := !ok || progress.Detail == `{"status":"success"}`
 		if shouldExecute {
```
laverya: What do you do if this is false?
Do we want to, you know, not execute the step?

Rob0h: no we're always going to "re-execute"

laverya: Because it looks like we execute either way

Rob0h: that's the intent. i'm removing to avoid confusion

PR #735, opened 2018-11-30
Ship unfork

marccampbell:
What I Did
------------
Add a new "unfork" command to Ship. This takes two parameters: a URL to a forked/copied Helm chart, and a URL to an upstream Helm chart. Ship will run (in headless mode) to create Kustomize patches from the fork, so that the upstream helm chart can be used directly again. The values.yaml from the fork will be used when running Helm template against both charts.

How I Did it
------------
Modified Ship to support a new, default Unfork lifecycle. The "ship unfork" command always invokes this. This only works with 2 a forked Helm chart today.  Used the existing CreateTwoWayStrategicMergePatch functionality.

How to verify it
------------
Run the integration tests. Or:

```
ship unfork --upstream https://github.com/helm/charts/tree/353ba5ef6467fd64035b7d5446df426f86d60153/stable/grafana https://github.com/replicatedhq/test-charts/tree/2687f40789200c03c80df4831a56e9af77b2e701/grafana-forked
```

Description for the Changelog
------------
Add `ship unfork` command to create patches from an existing fork.

------------

clipart20572


integration/lib.go
```diff
@@ -121,3 +136,64 @@ func CompareDir(expected, actual string, replacements map[string]string) (bool,
 
 	return true, nil
 }
+
+func prettyAndCleanJSON(data []byte, keysToIgnore []string) []byte {
```
ship-version-checker: do we still want to add this along with `modifyField`?

Rob0h: do we still want this and `modifyField` ?

pkg/lifecycle/unfork/daemonless.go
```diff
@@ -0,0 +1,148 @@
...
+	State   state.Manager
+	Patcher patch.ShipPatcher
+}
+
+func NewDaemonlessUnforker(logger log.Logger, fs afero.Afero, state state.Manager) lifecycle.Unforker {
+	return &Unforker{
+		Logger: logger,
+		FS:     fs,
+		State:  state,
+		Patcher: patch.ShipPatcher{
```
Rob0h: this can also be DI'd v.s. creating another instance

integration/lib.go
marccampbell: Removed `modifyField`. Left this pulled out of the `CompareDir` for now, it was repeated, so /shrug. It's probably better to be out?

PR #738, opened 2018-12-03
Add support for passing in helm values file

marccampbell:
What I Did
------------
Add support for passing `--helm-values-file` to ship init or update, which will overwrite the current values.yaml (from the upstream or saved). This is supported in headless mode right now.

How I Did it
------------
Add the flag, read the flag, read the file.

How to verify it
------------
Integration tests or run 

```
ship init github.com/helm/charts/stable/signalsciences --helm-values-file=./path/to/custom/values.yaml
```

Description for the Changelog
------------
- Add support for --helm-values-file to overwrite values.yaml in headless mode

------------


pkg/cli/root.go
```diff
@@ -72,6 +72,8 @@ func RootCmd() *cobra.Command {
 	cmd.PersistentFlags().String("resource-type", "", "upstream application resource type")
 	cmd.PersistentFlags().BoolP("prefer-git", "", false, "prefer the git protocol instead of using http apis")
 
+	cmd.PersistentFlags().StringP("helm-values-file", "", "", "Optional file path to Values.yaml to be used when rendering Helm charts")
```
Rob0h: could we specify here that this is specifically only for `headless` mode?

PR #740, opened 2018-12-03
add GOPATH/bin to path explicitly for go-bindata-assetfs

fshot:
What I Did
------------
Updated ship Makefile to enable building even if GOPATH/bin isn't present in user's path. With this change, we can remove some path-fiddling from the Ship Homebrew formula (which is one of the issues preventing our PR from getting merged).

How I Did it
------------
explicitly push GOPATH/bin into PATH before invoking go-bindata-assetfs

How to verify it
------------
remove GOPATH/bin from your path & 'make build'

Description for the Changelog
------------

------------

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -291,7 +291,7 @@ build_ship_integration_test:
 	docker build -t $(DOCKER_REPO)/ship-e2e-test:latest -f ./integration/Dockerfile .
 
 pkg/lifecycle/daemon/ui.bindatafs.go: .state/mark-ui-gitignored .state/build-deps web/app/.state/built-ui
-	go-bindata-assetfs -pkg daemon \
+	export PATH=$(GOPATH)/bin:$$PATH; go-bindata-assetfs -pkg daemon \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -291,7 +291,7 @@ build_ship_integration_test:
 	docker build -t $(DOCKER_REPO)/ship-e2e-test:latest -f ./integration/Dockerfile .
 
 pkg/lifecycle/daemon/ui.bindatafs.go: .state/mark-ui-gitignored .state/build-deps web/app/.state/built-ui
-	go-bindata-assetfs -pkg daemon \
+	export PATH=$(GOPATH)/bin:$$PATH; go-bindata-assetfs -pkg daemon \
```
divolgin: exporting seems like an overkill.   you could just run `$GOPATH/bin/go-bindata-assetfs`

fshot: I wish!  But go-bindata-assetfs calls go-bindata, which isn't found unless GOPATH/bin is in the path

PR #741, opened 2018-12-04
Update yaml dep for newline fix

Rob0h:
What I Did
------------
- Update yaml dep for newline fix

How I Did it
------------
- Update yaml dep for newline fix

How to verify it
------------
- `ship update` for istio or redis

Description for the Changelog
------------
- Update yaml dep for newline fix

------------
🛶

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -353,11 +353,11 @@
 
 [[projects]]
   branch = "improve-comment-parsing"
-  digest = "1:b19f021fce8ceef9284606de1375b68bde8ec3b4b72433ed724813195f4a12cb"
+  digest = "1:3cb2828c888836614087aaf50be0d5e511806ee8339d1731282cbfea746a5581"
   name = "github.com/emosbaugh/yaml"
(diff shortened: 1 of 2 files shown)


Gopkg.toml
```diff
@@ -124,3 +124,7 @@ ignored = [
 [[constraint]]
   branch = "master"
   name = "github.com/hashicorp/go-version"
+
+[[override]]
+  name = "github.com/Masterminds/sprig"
+  version = "v2.16.0"
```
laverya: Why is this added? It doesn't seem to change the lockfile

Rob0h: the first time i ran update it changed the sprig version back to `2.11.0`

laverya: And sprig was already updated to v2.16.0 in #737

laverya: That's odd...

laverya: @Rob0h why `override` then, and not `constraint`?

Overrides are something to be avoided when possible

PR #747, opened 2018-12-05
Support more use-cases for unfork

Rob0h:
What I Did
------------
- Support more use-cases for unfork

How I Did it
------------
- Only save the lists metadata from the forked upstream and use that in order to rebuild the rendered yaml
- Run splitting of list yaml and multidoc yaml for the upstream
- Account for k8s upstreams and return an error in the case the upstream is k8s and the fork is a helm chart
- Don't try to generate a patch for non-native k8s resources

How to verify it
------------
- Added integration tests

Description for the Changelog
------------
- Support more use-cases for unfork

------------
🛶


pkg/lifecycle/unfork/unforker.go
```diff
@@ -287,14 +292,28 @@ func (l *Unforker) generatePatches(fs afero.Afero, step api.Unfork, upstreamMap
 				return errors.Wrap(err, "read forked")
 			}
 
+			forkedResoruce, err := util.NewKubernetesResource(forkedData)
```
jadengore: `Resource`

pkg/lifecycle/unfork/unforker.go
```diff
@@ -287,14 +292,28 @@ func (l *Unforker) generatePatches(fs afero.Afero, step api.Unfork, upstreamMap
...
 			if err := yaml.Unmarshal(forkedData, &forkedMinimal); err != nil {
 				return errors.Wrap(err, "read forked minimal")
 			}
 
 			upstreamPath, exists := upstreamMap[forkedMinimal]
 			if !exists {
-				return errors.New(fmt.Sprintf("No matching upstream file found for %s", targetPath))
+				// If no equivalent upstream file exists, it must be a brand new file.
+				_, fileName := path.Split(relativePath)
+				overlay.Resources[string(filepath.Separator)+fileName] = string(forkedData)
```
jadengore: Somewhat hard to read this key, can you make this a variable?

pkg/lifecycle/unfork/unforker.go
laverya:
```suggestion
			forkedResource, err := util.NewKubernetesResource(forkedData)
```

laverya: We have suggested changes, might as well use them!

jadengore: @laverya has to be changed in more than one spot

jadengore: Also no IDE tools to guard against bad syntax

laverya: Awww

laverya: (but probably right)

PR #751, opened 2018-12-06
Use the metadata release name for k8s yaml from helm

Rob0h:
What I Did
------------
- Use the metadata release name for k8s yaml from helm

How I Did it
------------
- Move splitting of multi-doc yaml for cases where the fork is a single k8s file
- Read the release name from k8s yaml
- Use it during the helm templating of the upstream
- Use it during the update flow
- Exclude native k8s base resources that are not being used in the forked upstream

How to verify it
------------
- Tests, Integration

Description for the Changelog
------------
- Use the metadata release name for k8s yaml from helm

------------
🛶


integration/unfork/istio-k8s/metadata.yaml
```diff
@@ -1,3 +1,3 @@
 args: ["--prefer-git"]
 upstream: https://github.com/istio/istio/tree/1.0.3/install/kubernetes/helm/istio
-fork: https://github.com/replicatedhq/test-charts/blob/d05895fe4fc808b82a46a856dfda11e3b19feffb/istio-forked/rendered.yaml
+fork: https://github.com/replicatedhq/test-charts/blob/dc51b6517eb333d6ac7cc096c9a058dd278b1cee/istio-forked/rendered.yaml
```
Rob0h: didn't realize i hadn't merged this and i squashed and merged - hence the commit sha

pkg/specs/interface.go
```diff
@@ -348,6 +368,41 @@ func (r *Resolver) resolveUnforkRelease(
...
 }
 
+func (r *Resolver) maybeGetReleaseName(path string) (string, error) {
+	type k8sReleaseMetadata struct {
+		Metadata struct {
+			Labels struct {
+				Release string `yaml:"release"`
+			} `yaml:"labels"`
+		} `yaml:"metadata"`
+	}
```
jadengore: Any reason this type is defined in the function?

pkg/util/multidoc_yaml.go
```diff
@@ -7,25 +7,25 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
-	"github.com/replicatedhq/ship/pkg/util"
+	"github.com/spf13/afero"
 	yaml "gopkg.in/yaml.v2"
 )
 
 // this function is not perfect, and has known limitations. One of these is that it does not account for `\n---\n` in multiline strings.
```
jadengore: Does this comment still apply?

pkg/specs/interface.go
Rob0h: it's only used here to take out the release name

pkg/util/multidoc_yaml.go
Rob0h: yes

PR #758, opened 2018-12-12
16604 key commands

GraysonNull:
What I Did
------------
- Add handler for `cmd+s/ctl+s` keystrokes to save `values.yaml`
- Set initial value of editor through ace instead of just state to reset the UndoManager and prevent `cmd+z/ctl+z` from wiping out the entire editor.

How to verify it
------------
- On the values.yaml editor, make a change and do `cmd+s`. Your changes should be saved.
- Make additional changes and then hit `cmd+z` and bunch of times. It should undo your changes to the beginning but it should not ever wipe out the entire contents of the editor.

Description for the Changelog
------------
Support `CMD+S/CTL+S` to save your values.yaml
Bug fix for to many undo's wiping out entire editor.

------------
300px-uss_strong_ dd-758 _underway_off_oahu_in_1968

### DIFF
--- a/web/init/src/components/kustomize/HelmValuesEditor.jsx
+++ b/web/init/src/components/kustomize/HelmValuesEditor.jsx
@@ -31,7 +31,8 @@ export default class HelmValuesEditor extends React.Component {
   }
 
   componentDidMount() {
-    if(this.props.getStep.values) {
+    window.addEventListener("keydown", this.handleKeyboardSave);
+    if (this.props.getStep.values) {
       this.setState({
         initialSpecValue: this.props.getStep.values,
         specValue: this.props.getStep.values,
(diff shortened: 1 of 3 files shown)


web/init/src/components/kustomize/HelmValuesEditor.jsx
```diff
@@ -138,6 +142,18 @@ export default class HelmValuesEditor extends React.Component {
     }
   }
 
+  handleKeyboardSave = (e) => {
+    if (e.keyCode == 83 && (e.ctrlKey || e.metaKey)) {
```
jadengore: Would prefer if you name the keycode with a descriptive variable name as a constant

jadengore: Also `===`

PR #759, opened 2018-12-12
Fix UI for workflows with multiple async lifecycle steps

Rob0h:
What I Did
------------
- Fix UI for workflows with multiple async lifecycle steps

How I Did it
------------
- Previously, we were finding routes via the current `phase` that we are in (e.g. `kubectl`, `terraform`, etc.) Route phases are not unique as multiple `kubectl` steps will have the same `phase`. In order to reconcile this, the route `id` is now being used as it is unique to the each step.

How to verify it
------------
- Run `ship init fixtures/kubectl/ship.yml` to observe fixed functionality
- Previous functionality can be verified via `ship init fixtures/terraform/ship.yaml` and `ship init ${helm_chart}`

Description for the Changelog
------------
- Fix UI for workflows with multiple async lifecycle steps

------------
🛶

### DIFF
--- a/web/init/src/components/shared/StepBuildingAssets.jsx
+++ b/web/init/src/components/shared/StepBuildingAssets.jsx
@@ -5,12 +5,17 @@ import { Line } from "rc-progress";
 import { Utilities } from "../../utilities/utilities";
 import Loader from "./Loader";
 
-export default class StepBuildingAssets extends React.Component {
+export const RENDER_PHASE = "render";
+
+export class StepBuildingAssets extends React.Component {
   static propTypes = {
     location: PropTypes.shape({
       pathname: PropTypes.string,
     }).isRequired,
-    routeId: PropTypes.string.isRequired,
+    currentRoute: PropTypes.shape({
+      id: PropTypes.string,
+      phase: PropTypes.string,
+    }).isRequired,
     startPollingStep: PropTypes.func.isRequired,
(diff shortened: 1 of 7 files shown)


web/init/src/components/shared/StepBuildingAssets.jsx
```diff
@@ -19,8 +24,10 @@ export default class StepBuildingAssets extends React.Component {
   }
 
   componentDidMount() {
-    const { startPollingStep, location, routeId } = this.props;
-    startPollingStep(location, routeId);
+    const { startPollingStep, currentRoute } = this.props;
+    if (currentRoute.phase === RENDER_PHASE) {
```
Rob0h: This logic is here for these async steps to guard against re-polling. These components are re-rendered as the transition happens to the next phase as we're directly outputting JSX from the `renderStep` in `DetermineComponentForRoute`

web/init/src/components/shared/StepTerraform.spec.js
```diff
@@ -1,13 +1,16 @@
...
 import renderer from "react-test-renderer";
 
 const mockProps = {
   location: {
     pathname: "",
   },
-  routeId: "",
+  currentRoute: {
+    phase: TERRAFORM_PHASE,
+    id: "hotdog",
```
jadengore: 😂

Issue #762, opened 2018-12-12
ship watch leaks temp files


divolgin:
### Expected behavior

When watching a single repo, ship watch should only keep one git cloned copy of the repo.

### Actual behavior

Every time the timer fires, ship watch creates a new temp directory and abandons the previous one.

### Information

- Ship version: 0.28.0
- Docker version: 18.06.0 (but any will do)
- Command line run: see Steps to reproduce

### Steps to reproduce the behavior

1. Create a watch
    ```
    ./bin/ship init https://github.com/helm/charts/stable/grafana
    ./bin/ship watch --interval 5s
    ```
3. Let the watch run for a few polling intervals
4. Check output directory.  There should be only one copy of the repo, however:
    ```
    $ ls -l .ship/tmp/
    total 0
    drwxr-xr-x  15 user  staff  480 Dec 12 16:27 repo045526806
    drwxr-xr-x  15 user  staff  480 Dec 12 16:28 repo117690749
    drwxr-xr-x  15 user  staff  480 Dec 12 16:28 repo151529143
    drwxr-xr-x  15 user  staff  480 Dec 12 16:27 repo204546351
    drwxr-xr-x  15 user  staff  480 Dec 12 16:27 repo255347906
    drwxr-xr-x  15 user  staff  480 Dec 12 16:27 repo265968595
    drwxr-xr-x  15 user  staff  480 Dec 12 16:28 repo289046200
    drwxr-xr-x  15 user  staff  480 Dec 12 16:27 repo397560633
    drwxr-xr-x  15 user  staff  480 Dec 12 16:27 repo453984580
    ```

PR #766, opened 2018-12-13
Don't leak watch files

divolgin:
What I Did
------------

Fixed a temp file leak in watch.

How I Did it
------------

Delete the temp folder with git repo contents.

How to verify it
------------

Follow steps to reproduce in #762 

Description for the Changelog
------------

Fixed file leak when running `ship watch`.

------------

leak ship


pkg/specs/interface_test.go
```diff
@@ -209,18 +221,21 @@ icon: https://kfbr.392/x5.png
...
+				localApp *apptype.MockLocalAppCopy,
 				mockState *state.MockManager,
 				mockFs afero.Afero,
 				mockAppResolver *replicatedapp.MockResolver,
 				mockGitHubFetcher *githubclient.MockGitHubFetcher,
 			) {
 				inOrder := mockUi.EXPECT().Info("Reading /path/to/ship.yaml?customer_id=123&installation_id=456&release_semver=789 ...")
 				inOrder = mockUi.EXPECT().Info("Determining application type ...").After(inOrder)
-				inOrder = appType.EXPECT().
+				inOrder = inspector.EXPECT().
```
dexhorthy: nice

pkg/specs/interface.go
```diff
@@ -210,49 +207,48 @@ func (r *Resolver) ResolveRelease(ctx context.Context, upstream string) (*api.Re
...
+	app, err := r.appTypeInspector.DetermineApplicationType(ctx, upstream)
 	if err != nil {
 		return "", errors.Wrapf(err, "resolve app type for %s", upstream)
 	}
-	debug.Log("event", "apptype.inspect", "type", appType, "localPath", localPath)
+	debug.Log("event", "apptype.inspect", "type", app.GetType(), "localPath", app.GetLocalPath())
 
 	defer func() {
-		if err := r.FS.RemoveAll(localPath); err != nil {
+		if err := app.Remove(r.FS); err != nil {
```
dexhorthy: what is the reasoning behind passing in the FS here? I would have expected to just see `localPath` replaced with `app.GetLocalPath()` everywhere (which is a really solid refactor IMO). what does this get us over `r.FS.RemoveAll(app.GetLocalPath)`

dexhorthy: (not saying this is a blocker, just want to understand the motivations, and what problems you may have hit that I'm not thinking of)

divolgin: @dexhorthy `localPath` is the location of the chart within the cloned repo, but we need to remove the entire temp folder, which is some arbitrary number of directories up in the file tree.

divolgin: If you look in that `Remove` function, it doesn't remove localPath but a different path.

dexhorthy: gotchaaa okay makes sense. Thanks for clarifying

PR #770, opened 2018-12-18
Readme updates

fshot:
What I Did
------------
Expanded the readme with more specific usage examples, description of files written by ship, mentions of Ship Cloud and ship.yaml.

How I Did it
------------

How to verify it
------------

Description for the Changelog
------------

------------
bigbookofships

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ This is possible because of how the three operating modes of Ship invoke, store,
 
 # Three operating modes
 
+![Replicated Ship Modes](https://github.com/replicatedhq/ship/blob/master/logo/ship-flow.png)
+
 ## ship init
 Prepares a new application for deployment. Use for:
 - Specifying the upstream source for an application to be managed -- typically a repo with raw Kubernetes yaml or a Helm chart
@@ -33,8 +35,6 @@ Polls an upstream source, blocking until any change has been published.  Use for
 Updates an existing application by merging the latest release with the local state and overlays. Use for:
 - Preparing an update to be deployed to a third party application
 - Automating the update process to start from a continuous integration (CI) service
-
-![Replicated Ship Modes](https://github.com/replicatedhq/ship/blob/master/logo/ship-flow.png)
 
 # Features
 Ship is designed to provide first-time configuration UI and/or be used headless in a CI/CD pipeline to automate deployment of third party applications.
@@ -81,7 +81,7 @@ After Ship is installed, create a directory for the application you'll be managi
 ```shell
 mkdir -p ~/my-ship/example
 cd ~/my-ship/example
-ship init <path-to-chart> # github.com/helm/charts/stable/grafana
+ship init <path-to-chart> # github.com/helm/charts/tree/master/stable/grafana
 ```
 
 You'll be prompted to open a browser and walk through the steps to configure site-specific values for your installation, updating Helm values (if it's a chart), and making direct edits to the Kubernetes yaml (or Helm-generated yaml), which will be converted to patches to apply via Kustomize.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,59 +43,103 @@ Ship is designed to provide first-time configuration UI and/or be used headless
...
 - Merge [Helm](https://helm.sh/) charts with override values and apply custom overlays with [Kustomize](https://kustomize.io/) to avoid merge conflicts when upstream or local values are changed
 - Watch upstream repos for updates & sync changes to your local version.
-- Deploy [Helm](https://helm.sh/) charts to a Kubernetes cluster without Tiller
+- Render [Helm](https://helm.sh/) charts to pure Kubernetes yaml that can be deployed to clusters without Tiller
 - Enables [GitOps](https://www.weave.works/blog/the-gitops-pipeline) workflows to update third party applications
 - Configuration workflow `ship.yaml` files can be included in [Helm](https://helm.sh/) chart repos, to customize the initial `ship init` experience
 
-# Installation
-There are two ways you can get started with Ship:
+# Getting Started
```
dexhorthy: If this isn't going to be the first thing, I think we should consider a table of contents at the top of the readme.  I've used [this tool](https://github.com/ekalinin/github-markdown-toc) for generating a TOC in the past

README.md
```diff
@@ -40,59 +43,103 @@ Ship is designed to provide first-time configuration UI and/or be used headless
...
 
-Ship is also available through [Homebrew](https://brew.sh/):
+- ... or you can install with [Homebrew](https://brew.sh/):
 ```shell
 brew tap replicatedhq/ship
 brew install ship
 ```
 
-After ship is installed, run it with:
+Alternately, you can run Ship in Docker, but you will need to mount and configure a shared volume, in order to persist any changes made within the Ship admin console when launched via Docker.
```
dexhorthy: I think rather than (or in addition to) an explanation about a shared volume being required, we should just include the `-v $PWD:/out` in the example docker run command.

README.md
fshot: Good call, added a TOC

README.md
fshot: Yes indeed.  Examples updated

README.md
```diff
@@ -40,59 +43,103 @@ Ship is designed to provide first-time configuration UI and/or be used headless
...
-A simple, starting workflow could be to run `ship watch && ship update` after completing `ship init`.
-This will apply an update to the base directory.
+The `watch` command is a trigger for CI/CD processes, watching the upstream application for changes. Running `ship watch` will load the local state file (which includes a content hash of the most recently used upstream) and periodically poll the upstream application and exit when it finds a change. `ship update` will regenerate the deployable application assets, using the most recent upstream version of the application, and any local configuration from `state.json`.  The new `rendered.yaml` output can be deployed directly to the cluster, or submitted as a pull request into a 'GitOps' repo.
+
+You can see this flow in action by running `ship init <path-to-chart>` with a chart repo you have commit privileges on, then `ship watch --interval 10s && ship update` to start polling, then commit a change to the upstream chart and see the `ship watch` process exit, with rendered.yaml updated to reflect the change.
```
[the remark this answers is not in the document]
fshot: I like it.  Flows much better.

marccampbell:
@fshot Do you think we can get the same message without leading with that table of contents? The h1,h2,etc tags feel effective, and I worry that a README that requires a TOC is a sign that the README should be split into multiple docs. I'm not suggesting that ours is too long, but just that the table of contents isn't needed IMO.

dexhorthy:
I had suggested the TOC intially because often I want to jump straight to the code samples, and not read a big long preamble. The TOC feels a little heavyweight seeing it now, but a good compromise might be to have just a link near the top about `skip to [getting started](#getting-started)

dexhorthy:
(the alternative might be, as you noted, to move architecture and feature docs to their own md file)

PR #778, opened 2019-01-02
Resolve local helm dependencies

Rob0h:
What I Did
------------
- Resolve local helm dependencies

How I Did it
------------
- An upstream url is created given a local helm dependency + upstream

How to verify it
------------
- Tests, Integration

Description for the Changelog
------------
- Resolve local helm dependencies

------------
🛶


integration/unfork/gcloud-sqlproxy/metadata.yaml
```diff
@@ -1,3 +1,3 @@
 args: ["--prefer-git"]
 upstream: https://github.com/rimusz/charts/tree/44e2acc78cd7b621ca306661770828f32e9a2820/stable/gcloud-sqlproxy
-fork: https://github.com/replicatedhq/test-charts/tree/859cd8dbed96e952a2c6901392e6fe341e56bdf3/gcloud-sqlproxy-forked
+fork: https://github.com/replicatedhq/test-charts/tree/a0d1eefbf133fa2776164d0be2ba2a382811c16c/gcloud-sqlproxy-forked
```
laverya: Why's this changing? Not that I'm objecting, since the rest of the test seems to have stayed the same, but not something I expected to see

Rob0h: i didn't merge it and then squash and merged haha

pkg/util/github_test.go
```diff
@@ -233,3 +237,49 @@ func TestParseGithubURL(t *testing.T) {
...
+			},
+			want: "github.com/o/r/tree/master/",
+		},
+		{
+			name: "github tree with a subdir",
+			githubURL: GithubURL{
+				Owner:  "o",
+				Repo:   "r",
+				Ref:    "master",
+				Subdir: "something/maybe",
```
laverya: General comment on test cases (I don't think this needs a fix, just something I prefer):
IMO it doesn't make sense to have an input overlap between all test cases (like Owner/Repo/Ref does here) because then you're not actually testing that input (as hardcoding would work in that case)

Either things are explicitly shared and shouldn't be in the test case definition, or they shouldn't be the same across all cases.

PR #780, opened 2019-01-03
Fix missing state file path in error message on "ship update"

dexhorthy:
What I Did
------------

Fix missing state file path on "ship update". Previously the following was printed

```
There was an unexpected error! No state file found at , please run "ship init"
```

How I Did it
------------

Set a default in the pflags StringVar `state-file` to the value of `constants.StatePath`

How to verify it
------------

```
$ rm -rf .ship && ship update
```

should print

```
There was an unexpected error! No state file found at .ship/state.json, please run "ship init"

There was an error configuring the application. A debug log has been written to ".ship/debug.log", please include it in any support inquiries.
```

Description for the Changelog
------------

Fix missing state file path in error message on "ship update"

:boat:


dexhorthy:
I think integration test failures may be legitimate. need to investigate

PR #781, opened 2019-01-03
Add a way to edit the helm release namespace in the UI

Rob0h:
What I Did
------------
- Add a way to edit the helm release namespace in the UI

How I Did it
------------
- Add an input on the helm values screen in advanced options for namespace
- Stores the namespace in state - used during `helm template`

How to verify it
------------
- Tests, Integration

Description for the Changelog
------------
- Add a way to edit the helm release namespace in the UI

------------
🛶


web/init/src/components/shared/NavBar.jsx
```diff
@@ -126,7 +126,7 @@ export class NavBar extends React.Component {
             imageLoaded: true,
           })
         })
-        .catch(() => this.setState({ updatedState }))
+        .catch(() => this.setState({ ...updatedState, imageLoaded: true }))
```
jadengore: This is fine but it will completely disable retries.

Thanks for catching my bug here too 😄

PR #783, opened 2019-01-03
Save and continue with defaults, remove disabled attr on button, show errors instead

10Dimensional:
What I Did
------------
Made it so that a person can click through the config screen with just the default values.

How I Did it
------------
Removed unused code from `ConfigOnly.jsx` and and set the initial form item data in state in the same component. Passed `getData` method down to `ConfigRender.jsx` instead of defining it there. Save using both buttons but only continue if `nextStep = true`.

How to verify it
------------
Run ship init with an app that has a `ship.yaml` file, probably one from our test [repo](https://github.com/replicatedhq/test-charts/).

Description for the Changelog
------------
Save config with default values instead of needing to edit inputs

------------
⛵️


10Dimensional:
Would prefer @GraysonNull to look at this since I'm commenting out some code. Some of the config render stuff was a little confusing, want to make sure I didn't mess anything up.

PR #789, opened 2019-01-08
Remove failing unfork test case

fshot:
What I Did
------------
Removed the elastic-stack unfork integration test.

How I Did it
------------

How to verify it
------------

Description for the Changelog
------------

------------


marccampbell:
@fshot any reason we shouldn't pin this to the older sha that validates the functionality initially intended?

PR #791, opened 2019-01-08
Fix elastic-stack unfork integration

Rob0h:
What I Did
------------
Fix elastic-stack unfork integration

How I Did it
------------
- Add upstream and fork dependencies to test-charts

How to verify it
------------
- Passing integration

Description for the Changelog
------------
Fix elastic-stack unfork integration

------------
🛶


pkg/lifecycle/render/helm/dependency.go
```diff
@@ -21,38 +23,41 @@ func (f *LocalTemplater) addDependencies(
...
 				if err := f.Commands.RepoAdd(repoName, dependency.Repository, helmHome); err != nil {
-					return errors.Wrapf(err, "add helm repo %s", dependency.Repository)
+					return depPaths, errors.Wrapf(err, "add helm repo %s", dependency.Repository)
 				}
 			}
 		}
 	}
 
-	return nil
+	return depPaths, nil
```
Rob0h: returning the paths of the local dependencies fetched here so that they can be cleaned up later during `cleanUpAndOutputRenderedFiles`

PR #794, opened 2019-01-10
Add --preserve-state flag

Rob0h:
What I Did
------------
Add --preserve-state flag

How I Did it
------------
Add --preserve-state flag to preserve existing state and treat it like a `ship update --headed` lifecycle

How to verify it
------------
Integration

Description for the Changelog
------------
Add --preserve-state flag

------------
🛶


pkg/ship/kustomize.go
```diff
@@ -63,15 +65,29 @@ func (s *Ship) Init(ctx context.Context) error {
...
-			if err := s.promptToRemoveState(); err != nil {
-				debug.Log("event", "state.remove.prompt.fail")
-				return err
+			if removeExistingState {
+				if err := s.promptToRemoveState(); err != nil {
+					debug.Log("event", "state.remove.prompt.fail")
+					return err
+				}
+			} else {
+				s.UI.Info(fmt.Sprint("Preserving current state"))
```
jadengore: Is there any reason to wrap this in `fmt.Sprint`?

Rob0h: no, probably an artifact of old changes

PR #799, opened 2019-01-18
Docker image test

emosbaugh:
What I Did
------------
Added an integration test to test public and private docker images.

How I Did it
------------
Added two docker assets, one public the other pointing to an external registry integration to the test account that proxies to an image in the staging replicated registry qa account.

How to verify it
------------
Run the integration tests

Description for the Changelog
------------
None?

------------

dockers


integration/init_app/docker-asset/expected/.ship/release.yml
```diff
@@ -0,0 +1,16 @@
+---
+assets:
+  v1:
+    - docker:
+        image: busybox:latest
```
laverya: `replicated/helloscratch:1.0.0` is used elsewhere (in the `base` tests) and is only 12k instead of 1.3M
We could create a private version to use as well

dexhorthy: +1 on andrew's comment re: helloscratch, busybox:latest is gonna change a lot

emosbaugh: didnt know about it. good point

PR #805, opened 2019-01-25
Initial support for KubeSeal template function

### LINKED ISSUE #802: Built-in support for SealedSecrets?
This is a feature request/proposal to include [SealedSecrets](https://github.com/bitnami-labs/sealed-secrets) support natively in Ship.

Ship is built for and recommends a GitOps workflow. Secret management can be difficult in this process. There are several ways that folks can manage secrets when using GitOps:

1) Manage secrets out of band. This might include using a specific tool such as Hashicorp Vault or a cloud providers KMS. Alternatively, out-of-band secret management could simply be manually deploying secrets using `kubectl` to prevent them from being commited.

2) Check the secrets in to source control. This is not advisable, even in a private repo because secrets often contain sensitive data and should not be shared in a git repo.

3) Encrypting the secrets and committing only the encrypted secret to the GitOps repo. This is possible with an operator such as Bitnami's SealedSecret controller. 

Ship, when using a `ship.yaml` file, could have native, built-in support for managing these better. Today, a cluster operator is expected to bring both an opinion and knowledge about how to mange secrets. Instead, Ship could recommend one of the solutions above and have built-in, native support for the workflow.

To kickstart this conversation, I'm proposing that Ship have native support for the SealedSecrets controller. We could have support for this in a way that does not require the `kubeseal` binary to be installed, but instead it could be vendored in or implemented in Ship directly.

This would be useful to prompt for a SealedSecret cert, and use it to create SealedSecret objects. For example, in a ship.yaml, an asset could look like:

```yaml
assets:
  v1:
    - inline:
        dest: ./sealedsecret.yaml
        contents: |
          ---
          apiVersion: bitnami.com/v1alpha1
          kind: SealedSecret
          metadata:
            name: my-secret
            namespace: '{{repl ConfigOption "namespace"}}'
          spec:
            encryptedData:
              password: '{{repl KubeSeal (ConfigOption "sealed_secret_cert") (ConfigOption "namespace") (ConfigOption "password") }}'
              privateKey: '{{repl KubeSeal (ConfigOption "sealed_secret_cert") (ConfigOption "namespace") (ConfigOption "private_key") }}'
```

The Kubeseal template function would be invoked on the render step, and have this definition:

```go
func (ctx ConfigCtx) kubeSeal(cert string, namespace string, clearText string) string {
        ...
}
```

marccampbell:
What I Did
------------
Added initial support for [SealedSecrets](https://github.com/bitnami-labs/sealed-secrets) by implementing a template function that will KubeSeal a secret value. This closes #802. 

How I Did it
------------
Studied the kubeseal code to understand the encryption implementation there. Reimplemented the same, short and standard encryption implementation in Ship. Also added tests to cover some of the basic functionality (bad cert, valid cert) use cases, and confirm data is returned. 

How to verify it
------------
I pushed a publicly available ship.yaml that uses this. You can install and run this ship application with:

```
ship init github.com/shipapps/signal-sciences
```

For reference, the implementation in that ship.yaml is at https://github.com/shipapps/signal-sciences/blob/master/ship.yaml. Note, you'll have to look at the generated sealedsecret.yaml on the filesystem after manually verifying. This runbook doesn't yet support adding the sealedsecret resource as a new resource in the ship overlay.

Description for the Changelog
------------
Added a KubeSeal template function to support natively creating SealedSecrets

------------

### DIFF
--- a/pkg/templates/config_context.go
+++ b/pkg/templates/config_context.go
@@ -10,11 +10,7 @@ import (
 	"github.com/replicatedhq/libyaml"
 )
 
-func (bb *BuilderBuilder) NewConfigContext(
-	configGroups []libyaml.ConfigGroup,
-	templateContext map[string]interface{},
-) (*ConfigCtx, error) {
-
+func (bb *BuilderBuilder) NewConfigContext(configGroups []libyaml.ConfigGroup, templateContext map[string]interface{}) (*ConfigCtx, error) {
 	builder := bb.NewBuilder(
 		bb.NewStaticContext(),
 	)
--- a/pkg/templates/ship_context.go
+++ b/pkg/templates/ship_context.go
@@ -28,7 +28,7 @@ func (bb *BuilderBuilder) NewShipContext() (*ShipContext, error) {
 	return shipCtx, nil
 }
 
-// FuncMap represents the available functions in the ConfigCtx.
+// FuncMap represents the available functions in the ShipCtx.
 func (ctx ShipContext) FuncMap() template.FuncMap {
 	return template.FuncMap{
 		"AmazonEKS": ctx.amazonEKS,
--- a/pkg/templates/ship_context_test.go
+++ b/pkg/templates/ship_context_test.go
@@ -36,7 +36,6 @@ func makeTestShipCtx(t *testing.T) ShipContext {
 
 // tests that generated CAs can be referenced again + that more than one CA can be generated
 func TestShipContext_makeCa_repeat(t *testing.T) {
-
 	tests := []struct {
 		name         string
 		caName       string
@@ -123,7 +122,6 @@ func TestShipContext_makeCa_repeat(t *testing.T) {
 // generates two certs, covering different sets of hosts, with the same CA
 // certs must cover the specified domains, be of the desired type, and be signed by the CA
 func TestShipContext_makeCert_repeat(t *testing.T) {
-
 	tests := []struct {
 		name           string
 		firstCertName  string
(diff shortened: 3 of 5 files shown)


pkg/templates/config_context.go
```diff
@@ -121,3 +131,64 @@ func (ctx ConfigCtx) getConfigOptionValue(itemName string) (string, error) {
...
+
+	if len(certs) == 0 {
+		err := errors.New("unable to find cert in supplied cert data")
+		level.Error(ctx.Logger).Log("msg", "failed to parse cert", "err", err)
+		return "", err
+	}
+
+	pubKey, ok := certs[0].PublicKey.(*rsa.PublicKey)
+	if !ok {
+		err := errors.New("cert was not a public key")
```
dexhorthy: want to add the type here?


something like 

```go
		err := errors.Errorf("cert was not a public key, was %T", certs[0].PublicKey)
```

pkg/templates/config_context.go
```diff
@@ -65,6 +74,7 @@ func (ctx ConfigCtx) FuncMap() template.FuncMap {
 		"ConfigOptionData":      ctx.configOptionData,
 		"ConfigOptionEquals":    ctx.configOptionEquals,
 		"ConfigOptionNotEquals": ctx.configOptionNotEquals,
+		"KubeSeal":              ctx.kubeSeal,
```
dexhorthy: It seems a little unexpected to me that this exists on ConfigContext and not on StaticContext -- does it explicitly rely on config options?

Issue #832, opened 2019-02-18
Adding dependencies fails for alias-based repositories


joeyb:
### Expected behavior

Recent snapshots of Istio 1.1 contain dependency repository references that are aliases (e.g. https://github.com/istio/istio/blob/1.1.0-snapshot.6/install/kubernetes/helm/istio/requirements.yaml#L64). If those aliases aren't already set (via previous `helm repo add` calls), then ship should either prompt for the URLs for those aliases or fail with a useful error message (the latter is likely much easier, and would be good enough).

### Actual behavior

Regardless of whether or not you have a repo added with the given alias, ship fails on the 3rd step of `ship init` with the following error message:

```
execute render step: execute plan: execute templating: add requirements deps for https://github.com/istio/istio/tree/1.1.0-snapshot.6/install/kubernetes/helm/istio: add helm repo @istio.io: helm repo add failed, output "": Could not find protocol handler for:
```

It looks like the code in ship that handles those dependencies does not understand aliases: https://github.com/replicatedhq/ship/blob/57fe3a3ccddb5d6109400db950018ace52da25f4/pkg/lifecycle/render/helm/dependency.go#L33

That block needs an additional clause that handles repositories starting with `@` or `alias:` and treats them as alias references.

### Information

- Ship version: 0.33.0
- Docker version: 18.09.1
- Command line run: `ship init https://github.com/istio/istio/tree/1.1.0-snapshot.6/install/kubernetes/helm/istio`

### Steps to reproduce the behavior

1. `ship init https://github.com/istio/istio/tree/1.1.0-snapshot.6/install/kubernetes/helm/istio`
2. Leave the default values and hit next, should fail on the 3rd step with the error message described above.

dexhorthy:
Want to give a quick update on this issue with the state of discussion and some workarounds. It looks like there are two things at play here:

- missing alias parsing in the `if repoURL.Scheme == ... ` cases  (I believe @joeyb  may have a fix for this)
- Ship creates its own temporary HELM_HOME, which means that even if a user 

### Proposed solutions

Right now the proposed solutions to the helm home mismatch are

- Use the user's configured HELM_HOME, either automatically or via a flag (I believe we may have hit an issue with this in the past but I don't recall what it was)
- Parse HELM_HOME and try to merge the user's ~/.helm stuff into the temporary helm home created by Ship
- Add a flag to `ship init` that would allow a user to set helm aliases at runtime (we'd probably need to store these in state.json for updates)
[...]

PR #836, opened 2019-02-21
WIP: Support for helm repository aliases

joeyb:
Related to #832

Relevant comment from slack:

> I've been digging into https://github.com/replicatedhq/ship/issues/832 and the solution is going to be a bit more complicated than I thought since ship creates its own helm home. After adding basic support for aliases to the func mentioned in that issue, I can get it working with the following flow, but it's pretty cumbersome:

```
$ ship init https://github.com/istio/istio/tree/1.1.0-snapshot.6/install/kubernetes/helm/istio
# From another terminal, before proceeding to the ship UI
$ helm init --client-only --home ~/ship-istio/.ship/tmp/.helm
$ helm --home ~/ship-istio/.ship/tmp/.helm repo add istio.io https://storage.googleapis.com/istio-release/releases/1.1.0-snapshot.6/charts
```

> I suspect the right solution would be to add a step in the UI for configuring the URLs for any repo aliases found in the project's `requirements.yaml`, but that seems like a fairly large chunk of work. Maybe it's sufficient to just add a flag to `ship init` that lets you set any needed alias URLs?

This PR currently just contains the changes needed to make the `addDependencies` method not choke on aliased dependencies, but it does not solve the problem around more elegantly adding those aliases to ship's helm home.

### DIFF
--- a/pkg/lifecycle/render/helm/dependency.go
+++ b/pkg/lifecycle/render/helm/dependency.go
@@ -26,6 +26,10 @@ func (f *LocalTemplater) addDependencies(
 ) (depPaths []string, err error) {
 	for _, dependency := range dependencies {
 		if dependency.Repository != "" {
+			if strings.HasPrefix(dependency.Repository, "@") || strings.HasPrefix(dependency.Repository, "alias:") {
+				// The repository is an alias. Assume it has already been added.
+				return depPaths, nil
+			}
 			repoURL, err := url.Parse(dependency.Repository)
 			if err != nil {
 				return depPaths, errors.Wrapf(err, "parse dependency repo %s", dependency.Repository)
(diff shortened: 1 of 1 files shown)


dexhorthy:
Thanks for the contribution! I think this is definitely a step in the right direction! @laverya wanna take a look?

pkg/lifecycle/render/helm/dependency.go
```diff
@@ -26,6 +26,10 @@ func (f *LocalTemplater) addDependencies(
 ) (depPaths []string, err error) {
 	for _, dependency := range dependencies {
 		if dependency.Repository != "" {
+			if strings.HasPrefix(dependency.Repository, "@") || strings.HasPrefix(dependency.Repository, "alias:") {
+				// The repository is an alias. Assume it has already been added.
+				return depPaths, nil
```
laverya: Could we `continue` here instead, so that any other dependencies that we do have the ability to resolve get added?

joeyb: ahh, definitely. i misread the method signature. that should definitely `continue` rather than returning.

joeyb: fixed

PR #879, opened 2019-03-22
FIx a typo in root.go

mumoshu:
What I Did
------------

Fix a typo in the `ship`'s help output, that is `laod` changed to `load` according to the context.

How I Did it
------------

Edited the root.go in GitHub :)

How to verify it
------------

Rebuild and run `ship` from the branch.

Description for the Changelog
------------

Fixed a typo in `ship`'s help output

------------

image


mumoshu:
One of CI builds has failed. But is it really caused by my change? 🤔 The error code `403` causing the failure seems to say no.

dexhorthy:
Hi @mumoshu -- thanks for this contribution. That test has a dependency on github, and as the test suite expands we've started to get rate limited, resulting in the 403 you saw.

It shouldn't be a blocker to get this merged.

PR #880, opened 2019-03-23
support for passing readme in values yaml

dexhorthy:
What I Did
------------

Allow app maintainers to override the readme shown in a `helmValues` step.

```yaml

lifeycle:
  v1:
    - render: {}
    - helmValues:
        path: installer/chart/valyes.yaml
        readme:
          contents: |
            ## Configuration guide for Some-Chart

            ...
```

Right now the UI reads this from the `/metadata` endpoint, but now a mantainer can override this.

How I Did it
------------

- add `readme.contents` as an optional field to `lifecycle.v1.helmValues`
- if `readme` is non null on the step, set the values of `daemontypes.HelmValues.Readme` when returning it down to the frontend.
- Frontend checks for a `readme` field on the step, otherwise uses a `readme` from the chart meta if present.

How to verify it
------------

Use a lifecycle like the following, and navigate to helm values step:

```yaml

lifeycle:
  v1:
    - render: {}
    - helmValues:
        path: installer/chart/valyes.yaml
        readme:
          contents: |
            ## Configuration guide for Some-Chart

            ...
```

Description for the Changelog
------------

Allow app maintainers to override the readme shown in a `helmValues` step.

:ship:

### DIFF
--- a/web/init/src/components/shared/Linter.jsx
+++ b/web/init/src/components/shared/Linter.jsx
@@ -4,12 +4,13 @@ import get from "lodash/get";
 
 export default class Linter extends React.Component {
 
-  constructor() {
-    super();
-    this.state = {
-      showHelp: false,
-      showPreview: true
-    };
+  constructor(props) {
+    super(props);
+
+    // if there's a readme, default to readme tab. Else default to linter results.
+    const showHelp = !this.props.readme;
+    const showPreview = !showHelp;
+    this.state = { showHelp, showPreview };
   }
 
   maybeLineNumber = (error) => {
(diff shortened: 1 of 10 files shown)


web/init/src/components/shared/Linter.jsx
```diff
@@ -4,12 +4,13 @@ import get from "lodash/get";
...
-      showHelp: false,
-      showPreview: true
-    };
+  constructor(props) {
+    super(props);
+
+    // if there's a readme, default to readme tab. Else default to linter results.
+    const showHelp = !this.props.readme;
+    const showPreview = !showHelp;
+    this.state = { showHelp, showPreview };
```
dexhorthy: This was not directly related but I don't think it makes sense to default to the readme tab if its empty (which happens in non init-chart workflows).

PR #884, opened 2019-03-26
templating: fallbacks to ease migration from console-flavored ship

dexhorthy:
What I Did
------------

- In `{{repl Installation "installation_id"}}`, if `license_id` was used to fetch, template in the license id value
- In `{{repl Installation "license_id"}}`, if `installation_id` was used to fetch, template in the installation id value
- Add `{{repl LicenseFieldValue "FIELD_KEY"}}` as an alias for `{{repl EntitlementValue "FIELD_KEY"}}

How to verify it
------------

- ship init replicated.app/slug?license_id=..., ensure `{{repl
Installation "installation_id"}} emits the given license ID.

and so on

Description for the Changelog
------------

- In `{{repl Installation "installation_id"}}`, if `license_id` was used to fetch, template in the license id value
- In `{{repl Installation "license_id"}}`, if `installation_id` was used to fetch, template in the installation id value
- Add `{{repl LicenseFieldValue "FIELD_KEY"}}` as an alias for `{{repl EntitlementValue "FIELD_KEY"}}

migrate the vendors

### DIFF
--- a/pkg/templates/installation_context.go
+++ b/pkg/templates/installation_context.go
@@ -16,24 +16,26 @@ type InstallationContext struct {
 	Logger log.Logger
 }
 
+func (ctx *InstallationContext) entitlementValue(name string) string {
+	if ctx.Meta.Entitlements.Values == nil {
+		level.Debug(ctx.Logger).Log("event", "EntitlementValue.empty")
+		return ""
+	}
+
+	for _, value := range ctx.Meta.Entitlements.Values {
+		if value.Key == name {
+			return value.Value
+		}
+	}
+
+	level.Debug(ctx.Logger).Log("event", "EntitlementValue.notFound", "key", name, "values.count", len(ctx.Meta.Entitlements.Values))
+	return ""
+}
+
 func (ctx *InstallationContext) FuncMap() template.FuncMap {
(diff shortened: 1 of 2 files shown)


pkg/templates/installation_context.go
```diff
@@ -16,24 +16,26 @@ type InstallationContext struct {
...
 }
 
-func (ctx *InstallationContext) FuncMap() template.FuncMap {
-	return template.FuncMap{
-		"EntitlementValue": func(name string) string {
-			if ctx.Meta.Entitlements.Values == nil {
-				level.Debug(ctx.Logger).Log("event", "EntitlementValue.empty")
-				return ""
-			}
+func (ctx *InstallationContext) entitlementValue(name string) string {
```
dexhorthy: weird whitespace things happening here because I pulled this up to a standalone function instead of [inline](https://github.com/replicatedhq/ship/pull/884/files#diff-9c37f1a6dbeb3cee12d6bd0ce83693c7L21).

marccampbell:
Nice. Since "installation_id" and "license_id" are actually the same value, this is good. The work we did to rename the field should have left the old names as alias. 

Thanks for adding!

PR #908, opened 2019-04-11
Adding license and more release info

divolgin:
What I Did
------------
Adding more license and release info

How I Did it
------------
By getting it from GraphQL service and saving it state file

How to verify it
------------
Look at state file

Description for the Changelog
------------
License release, assignee, channel, etc and release sequence are now available in release metadata

------------

### DIFF
--- a/pkg/state/models.go
+++ b/pkg/state/models.go
@@ -3,6 +3,7 @@ package state
 import (
 	"bytes"
 	"fmt"
+	"time"
 
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/replicatedhq/ship/pkg/api"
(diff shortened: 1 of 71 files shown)


pkg/state/models.go
```diff
@@ -90,8 +91,17 @@ type V1 struct {
 	Certs map[string]util.CertType `json:"certs,omitempty" yaml:"certs,omitempty" hcl:"certs,omitempty"`
 }
 
+type License struct {
```
marccampbell: This struct is duplicated, intentional? It's in pkg/state and pkg/api.

divolgin: I'll double check.. I was following the existing pattern.

PR #920, opened 2019-04-23
Ship configitem type file - styling

jgruica:
What I Did
------------
Added styling to ship config item type file

How I Did it
------------
Refactored and changed <FileInput /> component

How to verify it
------------
Create a ship yaml that contains in your config section
```
 - name: tls_cert
    title: TLS Cert
    type: file
    required: true
```
and new style will be applied

Description for the Changelog
------------
Added styling to ship config item type file

------------
image


web/init/src/components/config_render/FileInput.jsx
```diff
@@ -46,32 +55,43 @@ export default class FileInput extends React.Component {
...
-
     let label;
-    if (this.props.label) {
-      label = this.props.label;
-    } else if (this.props.multiple) {
-      label = "Choose files";
-    } else {
-      label = "Choose file";
-    }
+    this.props.label ? label = this.props.label : this.props.multiple
```
GraysonNull: Any particular reason for moving from what we had to this notation? In my opinion what we had before was easier to read.

web/init/src/components/config_render/FileInput.jsx
```diff
@@ -46,32 +55,43 @@ export default class FileInput extends React.Component {
...
+        <div className={`${this.props.readonly ? "readonly" : ""} ${this.props.disabled ? "disabled" : ""}`}>
+          <p className="sub-header-color field-section-sub-header u-marginTop--small u-marginBottom--small u-marginTop--15">{label}</p>
+          <div className="flex flex-row">
+            <div className={`${this.state.fileAdded ? "file-uploaded" : "custom-file-upload"}`}>
+              <input
+                ref={(file) => this.file = file}
+                type="file"
+                name={this.props.name}
+                className="inputfile"
+                id={`${this.props.name} "selector"`}
```
GraysonNull: Does `selector` need to be in quotes here? (and also on the label below it)

jgruica: doesn't need to be, fixing!

web/init/src/components/config_render/FileInput.jsx
jgruica: Changed it because it looks cleaner for me, but I can add it back how was before :)

web/init/src/components/config_render/FileInput.jsx
GraysonNull: 👍

web/init/src/components/config_render/FileInput.jsx
GraysonNull: I mean it's a matter of preference really. I'm fine with the change, just was curious as to why you updated it

PR #948, opened 2019-05-23
Bj task 20111

MrBenJ:
What I Did
------------
I added a `max-height` rule to make sure the page isn't super long when someone is viewing a long README.md file in the Helm Chart editor.

How I Did it
------------
* Add a `max-height` value to the HelmChartInfo component to add a scrollbar for long README files
* Put a ternary on a new line so it's easier to compare the string values
* Removed an unnecessary anonymous function
* added the lovely [classnames](https://github.com/JedWatson/classnames) package to the `init` package.

How to verify it
------------
1. Create a new watch by adding a popular Helm chart URL like DataDog or Graphana
2. Go through the flow until you get to the editor page.

Description for the Changelog
------------
Fixed an issue where long README.md files on the helm chart editor page would make the page extremely long when viewing the README.md file with the YAML editor.

------------
this-is-a-ship-shipping-ship-shipping-shipping-ships-31872584

### DIFF
--- a/web/init/src/scss/components/kustomize/HelmValuesEditor.scss
+++ b/web/init/src/scss/components/kustomize/HelmValuesEditor.scss
@@ -1,4 +1,11 @@
-.AceEditor--wrapper.helm-values {
+.AceEditor--wrapper {
+  max-height: 68vh;
+
+  &.shorten {
+    max-height: 50vh;
+  }
+
+  &.helm-values {
     margin-bottom: 30px;
     overflow: auto;
     border: 1px solid #C4C8CA;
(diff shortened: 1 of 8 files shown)


web/init/src/scss/components/kustomize/HelmValuesEditor.scss
```diff
@@ -1,4 +1,11 @@
-.AceEditor--wrapper.helm-values {
+.AceEditor--wrapper {
+  max-height: 68vh;
```
GraysonNull: what happens if you have a really tall display? Would there just be lots of whitespace at the bottom?

My thinking as to how to solve this was looking at the flex properties and the wrappers around it so that no matter the height of the display it would flex to all available space.

MrBenJ:
**NOTE:** Force push was used to rebase the `cypress` fixes from #949

PR #949, opened 2019-05-23
Upgrade cypress to ^3.2.0

MrBenJ:
What I Did
------------
Stop Cypress e2e tests from failing

How I Did it
------------
Upgrade Cypress from version 3.1 to 3.2 to allow use of the `{end}` special character

How to verify it
------------
After installing dependencies, run `make cypress_base` in the root folder

Description for the Changelog
------------
* Upgraded Cypress ^3.2.0 for e2e tests

------------
hXuPFtx

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,11 +112,11 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - ship-e2e-setup-cy-3.1.0
+            - ship-e2e-setup-cy-3.2.0
       - run:
           # When updating Cypress version,
           # update cache key below
-          name: Install Cypress 3.1.0
+          name: Install Cypress 3.2.0
           working_directory: ~/ship/web/app
           command: npm install cypress@3.2.0
       - run:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -118,7 +118,7 @@ jobs:
           # update cache key below
           name: Install Cypress 3.1.0
```
GraysonNull: Should probably update this text to avoid confusion

web/app/cypress/Dockerfile
```diff
@@ -11,7 +11,7 @@ FROM cypress/browsers:node8.9.3-chrome73
 # Instead, pull binary directly
 # TODO: Verify checksum of binary
 # See https://github.com/cypress-io/cypress/issues/812
-RUN curl https://download.cypress.io/desktop/3.1.0?platform=linux64 -L -o cypress.zip
+RUN curl https://download.cypress.io/desktop/3.2.0?platform=linux64 -L -o cypress.zip
 RUN mkdir -p /Cypress/3.1.0
 RUN unzip -q cypress.zip -d /Cypress/3.1.0
```
GraysonNull: same with lines 15 and 16, I think this would fail to unzip the 3.2.0 version

PR #951, opened 2019-05-28
Bj optimize bundle size

MrBenJ:
What I Did
------------
Optimize bundle size

How I Did it
------------
Strip comments out of CSS, remove unused dependencies, use UglifyJS to force everything into one chunk, and make everything as tiny and bundled as possible.

How to verify it
------------
In the `init` folder, run `SHIP_SHOW_BUNDLE_ANALYZER=true yarn build` and note the 500kb gzipped sized file.

Description for the Changelog
------------
* Decreased bundle size significantly

------------

StarWarsXWingStarfighterYellow3dsmodel01 jpg3F48FDB7-7EB6-4F79-8081-F3A123BE5336Original


MrBenJ:
That's because the `postcss.config.js` contents now live inside of the `postcss-loader` options inside of `webpack.config.js`. It won't search for an external config if the config is in the Webpack.

GraysonNull:
Gotcha, make sense thanks for clarifying!

PR #980, opened 2019-06-19
Pact tests for ship

salahalsaleh1993:
What I Did
------------
Created pact tests for the following queries:
1- GetSlugRelease
2- GetLicense

How I Did it
------------
1- created two test files, `get_slug_release_test.go` and `get_license_test.go`.
2- created the test
3- ran it
4- verified it

How to verify it
------------
`make pacts` , and look at the output file in `pacts` directory.

Description for the Changelog
------------
Added contract tests for GraphQL API Queries

------------

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -175,7 +175,7 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker
-      - run: make build-deps pacts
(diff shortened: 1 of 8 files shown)


laverya:
At some point in the future we should probably add some 'this is expected to fail' tests, but that can wait

PR #992, opened 2019-06-24
Making some watch functionality usable as a stand-alone component

divolgin:
What I Did
------------
Refactored code to make `MaybeResolveVersionedUpstream` and `ReadContentSHAForWatch` easily reusable.

How I Did it
------------

How to verify it
------------

Description for the Changelog
------------

------------


pkg/specs/content.go
```diff
@@ -0,0 +1,227 @@
...
+// stored in state is greater than the latest version. All other upstreams will be returned unmodified.
+func (c *ContentProcessor) MaybeResolveVersionedUpstream(ctx context.Context, upstream string, existingState state.State) (string, error) {
+	debug := level.Debug(log.With(c.Logger, "method", "resolveVersionedUpstream"))
+
+	if !util.IsGithubURL(upstream) {
+		return upstream, nil
+	}
+
+	debug.Log("event", "resolve latest release")
+	latestReleaseVersion, err := c.GitHubFetcher.ResolveLatestRelease(ctx, upstream)
```
laverya: I know this would be a logic change, but perhaps we should skip checking for a newer release via github if the persisted URL doesn't contain the UpstreamVersionToken?

pkg/specs/githubclient/client.go
```diff
@@ -18,7 +18,8 @@ import (
 	"github.com/google/go-github/github"
 	"github.com/pkg/errors"
 	errors2 "github.com/replicatedhq/ship/pkg/util/errors"
-	"github.com/spf13/afero"
+
+	// "github.com/spf13/afero"
```
laverya: Should this be removed entirely?

pkg/specs/githubclient/client_test.go
```diff
@@ -61,23 +61,21 @@ var _ = Describe("GithubClient", func() {
...
 			It("should fetch and persist README.md and Chart.yaml", func() {
 				validGitURLWithPrefix := "http://www.github.com/o/r/"
-				mockFs := afero.Afero{Fs: afero.NewMemMapFs()}
 				gitClient := &GithubClient{
 					Client: client,
-					Fs:     mockFs,
 					Logger: log.NewNopLogger(),
 				}
 
 				dest, err := gitClient.GetFiles(context.Background(), validGitURLWithPrefix, constants.HelmChartPath)
```
laverya: Should these be cleaned up after the fact? I think they should if we're not using a mock fs

pkg/specs/githubclient/client.go
divolgin: i removed afero from this package during the initial attempt, but ended up keeping it elsewhere, so I put it back.

pkg/specs/content.go
laverya: This should be a seperate PR if it ends up being desirable

PR #994, opened 2019-06-24
Support saving/loading state data using an HTTP URL

divolgin:
What I Did
------------
Added `url` value for `state-from` command line argument.

How I Did it
------------
Implemented the support.

How to verify it
------------
Run ship with new arguments

```
ship init --state-from=url --state-put-url http://somehost/path --state-get-url http://somehost/path
```

Description for the Changelog
------------
Command line argument `state-from` can have value `url`

------------

### DIFF
--- a/pkg/state/manager.go
+++ b/pkg/state/manager.go
@@ -1,11 +1,7 @@
 package state
 
 import (
-	"encoding/json"
 	"fmt"
-	"os"
-	"path/filepath"
-	"strings"
 	"sync"
 
 	"github.com/go-kit/kit/log"
@@ -19,9 +15,6 @@ import (
 
 	"github.com/spf13/afero"
 	"github.com/spf13/viper"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
 )
 
 type Manager interface {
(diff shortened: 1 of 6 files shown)


pkg/state/manager.go
```diff
@@ -50,6 +43,11 @@ type Manager interface {
 	AddCA(name string, newCA util.CAType) error
 }
 
+type StateSerializer interface {
```
laverya: Does this need to be public? It does not appear to be used elsewhere

divolgin: Nope. Does not need to be public at all.

pkg/state/state_secret.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		return State{}, errors.Wrap(err, "get in cluster config")
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return State{}, errors.Wrap(err, "get kubernetes client")
+	}
+
+	if s.secretNS == "" {
```
dexhorthy: it looks like the secret serializer checks if all required parameters are set. Do we check that the get/put URLs are provided when state-from=url ?

divolgin: I kind of debated this.  There is no viper object here, so it's not really appropriate to hardcode command line arguments here.  And you still get a pretty meaningful error if the URL is an empty string.

dexhorthy: I think agree: in either case, doing validation inside the serializer implementation seems like the wrong place. Feels like if we add validation, it should probably be a bit higher up, where we pull stuff out of viper

PR #1007, opened 2019-06-27
Cache state to improve performance with URL-based storage.

divolgin:
What I Did
------------
Improved performance when using HTTP-based backend for state storage.

How I Did it
------------
State is now loaded and persisted once when using secret and URL-based backends.  File based storage is still updated continuously.  State is now served from an in-memory cache, and state manager is now implemented as a singleton.

How to verify it
------------
Run ship init and ship update.

Description for the Changelog
------------
Improved performance for HTTP-based state storage.

------------

### DIFF
--- a/pkg/ship/kustomize.go
+++ b/pkg/ship/kustomize.go
@@ -27,7 +27,7 @@ func (s *Ship) InitAndMaybeExit(ctx context.Context) error {
 func (s *Ship) stateFileExists(ctx context.Context) bool {
 	debug := level.Debug(log.With(s.Logger, "method", "stateFileExists"))
 
-	existingState, err := s.State.TryLoad()
+	existingState, err := s.State.CachedState()
 	if err != nil {
 		debug.Log("event", "tryLoad.fail")
 		return false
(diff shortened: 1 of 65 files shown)


pkg/ship/kustomize.go
```diff
@@ -77,16 +81,16 @@ func (s *Ship) Init(ctx context.Context) error {
...
+	// if removeExistingState {
+	// 	if err := s.maybeWriteStateFromFile(); err != nil {
+	// 		return err
+	// 	}
+	// } else {
+	// 	debug.Log("event", "reset steps completed for existing state")
+	// 	if err := s.StateManager.ResetLifecycle(); err != nil {
+	// 		return errors.Wrap(err, "reset state.json completed lifecycle")
+	// 	}
+	// }
```
laverya: I'm not positive this can be removed
Does `ship update --headed` still work?

pkg/state/manager.go
```diff
@@ -247,7 +308,11 @@ func (m *MManager) getStateSerializer() (stateSerializer, error) {
 
 	switch stateFrom {
 	case "file":
-		return newFileSerializer(m.FS, m.Logger), nil
+		// statePath := m.V.GetString("state-file")
+		// if statePath == "" {
+		// 	statePath = constants.StatePath
+		// }
```
laverya: extra comments?

pkg/state/manager.go
```diff
@@ -303,15 +380,17 @@ func (m *MManager) SaveKustomize(kustomize *Kustomize) error {
...
-	err := m.FS.RemoveAll(statePath)
-	if err != nil {
-		return errors.Wrap(err, "remove state file")
-	}
+	// err = s.Remove()
+	// if err != nil {
+	// 	return errors.Wrap(err, "remove state file")
+	// }
+
+	m.cachedState = &State{}
```
laverya: Should this call the newly added `Remove` function too?

Also, commented code above

divolgin: good catch. this was just for debugging.

PR #1025, opened 2019-07-11
Added entitlementSpec and configSpec to state.json

salahalsaleh1993:
What I Did
------------
Query entitlement spec and config spec to be included in state.json to be used for the Config tab instead of displaying the state.json as raw data, and to be used in the to-be-added Entitlements tab in Ship Cluster 

How I Did it
------------

How to verify it
------------

Description for the Changelog
------------

------------

### DIFF
--- a/contracts/replicatedapp/get_slug_release_test.go
+++ b/contracts/replicatedapp/get_slug_release_test.go
@@ -77,8 +77,53 @@ func Test_GetSlugRelease(t *testing.T) {
 						"spec": dsl.Like(dsl.String("assets:\n  v1:\n    - inline:\n        contents: |\n          #!/bin/bash\n          echo \"installing nothing\"\n          echo \"config option: {{repl ConfigOption \"test_option\" }}\"\n        dest: ./scripts/install.sh\n        mode: 0777\n    - inline:\n        contents: |\n          #!/bin/bash\n          echo \"tested nothing\"\n          echo \"customer {{repl Installation \"customer_id\" }}\"\n          echo \"install {{repl Installation \"installation_id\" }}\"\n        dest: ./scripts/test.sh\n        mode: 0777\nconfig:\n  v1:\n    - name: test_options\n      title: Test Options\n      description: testing testing 123\n      items:\n      - name: test_option\n        title: Test Option\n        default: abc123_test-option-value\n        type: text\nlifecycle:\n  v1:\n    - render: {}\n")),
 						"images": []string{},
 						"githubContents": []string{},
+						"configSpec": dsl.Like(dsl.String("v1:\n  - name: test_options\n    title: Test Options\n    description: testing testing 123\n    items:\n      - name: test_option\n        title: Test Option\n        default: abc123_test-option-value\n        type: text\n")),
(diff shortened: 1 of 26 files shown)


pkg/specs/replicatedapp/graphql.go
```diff
@@ -25,7 +25,8 @@ const ShipRelease = `
     channelIcon
     semver
     releaseNotes
-    spec
+		spec
+		configSpec
```
laverya: indentation is broken here
I think you might have used tabs?

pkg/specs/replicatedapp/graphql.go
```diff
@@ -43,7 +44,8 @@ const ShipRelease = `
         size
         data
       }
-    }
+		}
+		entitlementSpec
```
laverya: indentation is broken here too

contracts/replicatedapp/get_slug_release_test.go
```diff
@@ -77,8 +77,53 @@ func Test_GetSlugRelease(t *testing.T) {
...
 						"images": []string{},
 						"githubContents": []string{},
+						"configSpec": dsl.Like(dsl.String("v1:\n  - name: test_options\n    title: Test Options\n    description: testing testing 123\n    items:\n      - name: test_option\n        title: Test Option\n        default: abc123_test-option-value\n        type: text\n")),
+						"entitlementSpec": dsl.Like(dsl.String("---\n- name: Has Contract\n  key: has_contract\n  type: boolean\n  default: false\n- name: Is SAML Allowed\n  key: is_saml_allowed\n  type: boolean\n  default: false\n- name: Has Private Support\n  key: is_support_ticket_allowed\n  type: boolean\n  default: false\n")),
 						"entitlements": map[string]interface{}{
-							"values": []string{},
+							"values": [3]map[string]interface{}{
```
laverya: Make sure that the provider state has this

salahalsaleh1993: i updated the fixtures in vandoor and tested the new pact json file there

pkg/specs/replicatedapp/graphql.go
salahalsaleh1993: this is just on Github, dunno why

contracts/replicatedapp/get_slug_release_test.go
laverya: Ok
Well let's wait for that API change to make it to staging then

integration/init_app/integration_test.go
```diff
@@ -155,6 +155,8 @@ var _ = Describe("ship init replicated.app/...", func() {
 
 					ignoreEntitlementSig := map[string][]string{
 						".ship/state.json": {
+							"v1.upstreamContents.appRelease.configSpec",
+							"v1.upstreamContents.appRelease.entitlementSpec",
```
laverya: once again, the indentation here is wrong (and in `update`)

laverya: (and this was my view wrapping things - whoops)

PR #1068, opened 2019-09-21
Kustomize3

jbrette:
Just in case you plan to migrate to kustomize 3 code base:

What I Did
------------
- Remove the vendor folder
- Get the go.mod to point at the kustomize 3.x version of kustomize.

How I Did it
------------
- force the vendor directory to be recreate. go mod vendor
- Fix compilation issues to get make build to successed after replacing sig.k8s.io/kustomize by sigs.k8s.io/kustomize/v3
- consolidate the handling of kind/apiVersion in the kustomization.yaml (lots of change replace apiVersion: "" ...)

How to verify it
------------

- go test -v
- make build
-  ship init github.com/helm/charts/tree/master/stable/artifactory

Description for the Changelog
------------

------------

e00_22_07

e00_23_00


jbrette:
@laverya I think a lot of the failed integration tests are linked to the fact the kustomize 3.x does output the objects in the same order than kustomize 2.x (even so we activated the "legacy" order transformer).

Should the tests be updated or should a transformer build to match the exact same order ?

laverya:
This looks really nice! Is there any reason for the removed vendor dir, though?

Also, regarding integration test changes - it's really not the best system, but "expected changes after code changes" is what we've been doing so far.

The only other thing is the business logic changes to handle the new resources/bases/patches behaviors, but that's not something I'd expect you to do 😆 (And I can handle it in another PR)

jbrette:
@laverya I did remove the vendor directory because:
- it was containing the kustomize 2.0.3 code that we don't want anymore
- there is a bug in kustomize that makes that your kustomization.yaml looked wrong. Containing "apiversion" instead of "apiVersion". Using go.mod replace allowed us to point temporary to a fixed version of kubernetes
- There are two commits part of that PR.

jbrette:
Will not have enough time to complete the integration test changes.

laverya:
@jbrette, thank you for the PR - I'll start from this and fixup the tests myself if that's all right with you?

jbrette:
@laverya No problem, I leave the branch around. According to the last test, mainly the integration tests were failing and one task requiring a very specific version of a module in the vendor folder.

PR #1072, opened 2019-09-27
Fix textdata config issue

salahalsaleh1993:
What I Did
------------
Fix getData not passing to parent component issue

How I Did it
------------

How to verify it
------------

Description for the Changelog
------------

------------

### DIFF
--- a/web/init/src/ShipConfigRenderer.jsx
+++ b/web/init/src/ShipConfigRenderer.jsx
@@ -33,7 +33,8 @@ export class ShipConfigRenderer extends React.Component {
         <ConfigRender
           fieldsList={groupsList}
           fields={_groups}
-          handleChange={this.props.handleChange}
+          handleChange={this.props.handleChange || (() => { return; })}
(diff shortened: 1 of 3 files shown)


web/init/src/ShipConfigRenderer.jsx
```diff
@@ -33,7 +33,8 @@ export class ShipConfigRenderer extends React.Component {
         <ConfigRender
           fieldsList={groupsList}
           fields={_groups}
-          handleChange={this.props.handleChange}
+          handleChange={this.props.handleChange || (() => { return; })}
```
MrBenJ: Small nit - you can just do `() => {}` on the right side of `||` for these params

MrBenJ: `return undefined` will be implied for each of these since it's a empty function

salahalsaleh1993: that's how it was before the latest update, want to get the code to the same state it was before, going to take some time to test again :)

PR #1084, opened 2019-11-05
Kustomize3 and Helm 2.16

jbrette:
What I Did
------------

How I Did it
------------

How to verify it
------------

Description for the Changelog
------------

------------


.circleci/config.yml
```diff
@@ -283,6 +283,7 @@ jobs:
           tar -zxvf golang.tar.gz -C /tmp
           sudo rm -rf ${GOROOT}
           sudo mv /tmp/go ${GOROOT}
+      - run: GO111MODULE=on $GO mod vendor
```
laverya: what's the reasoning behind removing the vendor folder just to revendor everything as part of the CI/build process?

jbrette:
everything but the kitchen sink seems been dump in the dependency of kustomize lately, Also copy paste of the official kubernetes code has been mode which will most likely cause conflict with things like helm code.

PR #1091, opened 2019-12-10
make make build build

emosbaugh:
What I Did
------------
made it build ?


pkg/specs/apptype/determine_type.go
```diff
@@ -140,34 +140,32 @@ func (i *inspector) determineTypeFromContents(ctx context.Context, upstream stri
 	}
 
 	finalPath, err := fetcher.GetFiles(ctx, upstream, repoSavePath)
-	if err != nil {
```
laverya: why is this being dropped?

laverya: nvm, gosimple is catching it

PR #1097, opened 2020-02-11
Support when for config groups

salahalsaleh:
Addresses part of https://github.com/replicatedhq/kots/issues/897

Add `when` field support for config groups


web/init/src/components/config_render/ConfigGroup.jsx
```diff
@@ -142,7 +142,8 @@ export default class ConfigGroup extends React.Component {
 
   render() {
     const { item } = this.props;
-    if (!this.isAtLeastOneItemVisible()) return null;
+    const hidden = item && item.when === "false";
```
laverya: maybe order of operations here?
```suggestion
    const hidden = (item && (item.when === "false"));
```

salahalsaleh1993: javascript would check if `item` exists first then evaluate the right hand expression `item.when === "false"`

laverya: I know
Personally I feel the second is cleaner
But whatever code style we're using elsewhere is fine
It's JS, don't rely on my style sense lol

PR #1138, opened 2020-12-24
use node 12, not node 8, in CI

laverya:
What I Did
------------

How I Did it
------------

How to verify it
------------

Description for the Changelog
------------

------------

USS Steuben County (LST-1138)")


laverya:
This doesn't _fix_ the e2e_init test, but it at least lets it run instead of failing out of hand

salahalsaleh (approved):

